PR #23, opened 2015-07-17
backwards compatibility for Ubuntu 12.04

RussTedrake:


patmarion:
I'm going to close this and open a new pull request for the previous commit e32e9db only.

note drake-distro submodule would have to point at e32e9db, not master after this merge.

PR #34, opened 2015-07-29
Newest tabledemo for fixed base changes and EXOTica planning fix

iamwolf:
This PR includes
- mapping panel task panel for fixed base arms (hidden for humanoids)
- tabledemo task panel for fixed base, tested on robot
- fixed default parameter sync for tabledemo task panel
- merged in @MarcoCar and @mauricefallon newest additions for the humanoids in tabledemo and changed them to the new parameter sync as well (tested)
- expanded ImageBasedFitting to be able to turn off showing affordances - used in tabledemo from now on for example (to just show the camera feed/imageview)
- expanded RobotiqHanddriver to send to commands to trigger engaging and disengaging motors. These will likely not be used for the Robotiq, but we are using the Robotiq hand driver more as a general hand driver and needed two custom messages - is it okay to have it here and mid-term have a templated hand driver that is adapted? Right now we are handling every command that comes from the Robotiq hand driver in a translator and adapt it to other hands
- modified mappingpanel/mappingdemo a bit, now can load arbitrary file directly as "map" polydata element, rather than having to rename it after loading [important for continuing to run segmentation algorithms for example]
- fixed planning timing issues when using external / EXOTica planning after fixing EXOTica - now uses the ikPlanner parrameter options to control joint speed


patmarion:
+1

iamwolf:
Added new commits to:
- ``plannerPublisher`` and ``affordanceManager``: support attached collision objects besides rigid/global/stationary collision objects
- ``tabledemo``: also sync the affordance, not only the fitted object

patmarion:
+1 for the new commits.

by the way, how do you like using the affordance updater?  I think it works ok but it has some issues that make it sort of clunky.  Animating plans with attached affordances doesn't work well when there is a robot state stream. I can imagine some different designs that would make it work better and be more general.

src/python/ddapp/startup.py
```diff
@@ -450,6 +451,9 @@ def fitPosts():
     mappingDemo = mappingdemo.MappingDemo(robotStateModel, playbackRobotModel,
                     ikPlanner, manipPlanner, footstepsDriver, atlasdriver.driver, lHandDriver, rHandDriver,
                     perception.multisenseDriver, view, robotStateJointController, playPlans)
+    if useMappingPanel: #  if false, might break task panel TODO: handle
```
mauricefallon: Should you check these?

src/python/ddapp/startup.py
```diff
@@ -476,6 +480,8 @@ def fitPosts():
     taskPanels['Surprise'] = surpriseTaskPanel.widget
     taskPanels['Terrain'] = terrainTaskPanel.widget
     taskPanels['Table'] = tableTaskPanel.widget
+    if ikPlanner.fixedBaseArm: # 'Edinburgh' switch - this mapping currently was only tested and works on fixed base arms
+        taskPanels['Mapping'] = mappingTaskPanel.widget
```
mauricefallon: ... related, not very clean

src/python/ddapp/tabledemo.py
```diff
@@ -82,7 +83,7 @@ def __init__(self, robotStateModel, playbackRobotModel, ikPlanner, manipPlanner,
 
         self.constraintSet = []
 
-        self.reachDist = 0.07
+        self.reachDist = 0.12
```
mauricefallon: Is this number used by Valkyrie planning?

iamwolf: No, not anymore since we switched to using Marco's collision free reaching. I will change it nonetheless and add a commit in a bit

src/python/ddapp/startup.py
iamwolf: I fixed this, removed the comment and tested that it works for LWR, Val2, and V5

src/python/ddapp/startup.py
iamwolf: Same as above - fixed and tested

iamwolf:
Great, thanks.
Any other comments / Anything else to consider/amend/test before this is ready for merging?

patmarion:
Looks good, ready to merge?

iamwolf:
@mauricefallon is this ready to merge?

mauricefallon:
I'm happy for it to be accepted.

PR #38, opened 2015-07-31
Sync hand joints (more general)

iamwolf:
Made syncing of hand joints more general. It checks whether the robot is any kind of Atlas and then keeps the old way (getRobotiqJoints) as it's underactuated and wouldnt return all hand joints from the hand model (getRobotiqJoints returns 1, 2, middle for each). Otherwise uses (all of) the hand model(s) to arrive at the hand joints, then filters out the base joints. 

I've compiled and started this up with v5, lwr, and val2. Then verified that ``teleopJointPropagator.jointIndices`` has indices. This is the case for v5 and lwr, and hence should work. Valkyrie doesnt seem to have hand models / factories defined yet.

cc @patmarion @mauricefallon

### DIFF
--- a/src/python/ddapp/startup.py
+++ b/src/python/ddapp/startup.py
@@ -362,14 +362,14 @@ if usePlanning:
 
     handJoints = []
     if drcargs.args().directorConfigFile.find('atlas') != -1:
-        handJoints = roboturdf.getRobotiqJoints()
+        handJoints = roboturdf.getRobotiqJoints() + ['neck_ay']
     else:
         for handModel in ikPlanner.handModels:
             handJoints += handModel.handModel.model.getJointNames()
         # filter base joints out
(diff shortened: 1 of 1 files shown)


patmarion:
Looks good to me.

Is neck_ay used by valkyrie or is that custom to atlas?  Perhaps the "if atlas" block should add robotiq + 'neck_ay' to create the full list of sync joints?

iamwolf:
Great point, moved it into the if-statement. 

Valkyrie uses two joints for the neck, but I haven't tested it in sync with the SCS simulator to sync unactuated joints yet [we might get all broadcasted, dont know for sure] - we need to investigate and perhaps add at a later date then if need be

patmarion:
By the way, another way to get existence of substrings is to use the "in" keyword instead of find():

<pre>
>>> 'base' in 'base_x'
True

>>> 'base' not in 'neck_ay'
True
</pre>

PR #40, opened 2015-08-03
turn valve in a human manner

mauricefallon:
Supports human-like valve turning:
* defaults to original finals-style turning but can use human style turning
* verified that both modes work with atlas v5 in the mit simulator
* demo with val in scs:
https://www.dropbox.com/s/fkyw1t70dfoyttd/valve_turning_with_val.ogv?dl=0

Other:
* disabled the point cloud re-centering (from greg)
* fixed a hard coded 'head' string

### DIFF
--- a/src/python/ddapp/roboturdf.py
+++ b/src/python/ddapp/roboturdf.py
@@ -30,6 +30,7 @@ with open(drcargs.args().directorConfigFile) as directorConfigFile:
     handCombinations = directorConfig['handCombinations']
     numberOfHands = len(handCombinations)
 
+    headLink = directorConfig['headLink']
 
 
 def getRobotGrayColor():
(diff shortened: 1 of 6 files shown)


src/python/ddapp/roboturdf.py
```diff
@@ -30,6 +30,7 @@
     handCombinations = directorConfig['handCombinations']
     numberOfHands = len(handCombinations)
 
+    headLink = directorConfig['headLink']
```
patmarion: This willl break config files that do not contain headLink.  I think that's ok.  That is how we have been operating so far.  Eventually we should talk about what must be in a config file and what is optional.  For example, to make this optional, we could say:

    headLink = directorConfig.get('headLink', '') # default to empty string if not defined by config file

mauricefallon: In this case 'head' means the direction the robot's main sensor is facing which all robots should have.
I'd prefer not to fail like this as its difficult to hunt down the failure.

But you're right about having a limited list of blocks in the cfg file.

PR #45, opened 2015-08-10
Fix affordance api call

patmarion:


patmarion:
note, slight update to affordance calls, fixes valve tests. cc @mauricefallon @iamwolf

patmarion:
btw, with the new change, it is no longer required to add uuid=newUUID() to the dictionary, the affordance manager will automatically insert it if it doesn't already exist

PR #49, opened 2015-08-12
add an install_prereqs script for compatibility with drake superbuild

rdeits:


rdeits:
@patmarion showed me where the existing script is

PR #51, opened 2015-08-14
Fix for constraint - only affects my system

cdarpino:


patmarion:
Why do you want to remove the fixed base constraint?

mauricefallon:
It think this will effect our operation in Edinburgh with a fixed based 'bot

Can you confirm @iamwolf ?

iamwolf:
From looking at the source code it should not as we use the flat ``fixedBaseArm`` rather than ``robotNoFeet``. There's an if not fixed base, else block above and this no-feet block simply removes all constraints again and then appends. It should not be called for fixed base robots (not verified in simulation/running)

cdarpino:
That is correct, I'm using robotNoFeet whereas you are using fixedBaseArm.
In any case, I realize I don't want to merge this! As Pat is saying this is the base, I thought I had the back two times, so closing this.

PR #52, opened 2015-08-14
Sdf loading vlad

mithrandir123:
Added option for loading affordances from a SDF file.
- Added a copy of pysdf from https://github.com/andreasBihlmaier/pysdf
- Removed dependency on ROS packages
- Bug fix: only first visual element and first collision element were loaded
- Created a loader class that parses the SDF and populates the scene with affordances based on collision objects defined in the file.
- Articulated structures (containing joints) are ignored.
- Only boxes and cylinders are loaded for now (because SCS can ontly use these shapes for contact detection).

### DIFF
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -141,6 +141,10 @@ set(python_files
   ddapp/thirdparty/NumericStringParser.py
   ddapp/thirdparty/numpyjsoncoder.py
   ddapp/thirdparty/transformations.py
+  ddapp/thirdparty/conversions.py
+  ddapp/thirdparty/geometry_msgs_mock.py
+  ddapp/thirdparty/naming.py
+  ddapp/thirdparty/pysdf.py
 
   urdf_parser_py/__init__.py
   urdf_parser_py/sdf.py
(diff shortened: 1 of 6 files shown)


patmarion:
This pull requests adds ddapp/SDF.py but it looks like it is unused.  Should that be removed?

I see you added a copy of tranformations.py into pysdf.  The director already contains this module, it can be imported from:

from ddapp.thirdparty import transformations


It would be helpful if your commit history showed the modifications that you made to pysdf.   That way if we needed to update pysdf it would be clear what changes need to be ported on top.

It looks like to combined the pysdf files into one.  What was the motivation for that?  Also, maybe it's possible to add pysdf to ddapp/thirdparty?

mithrandir123:
The ddapp/SDF.py was just a temporary file. It has been removed now.

I have moved the pysdf library into ddapp/thirdparty and imported transformations from the same directory. pysdf can be imported from:
```Python
from ddapp.thirdparty import pysdf
```

I have included the change history and split the monolithic file into the 4 files it was composed of originally.

patmarion:
nice!

PR #58, opened 2015-08-19
Fixed SDF shape transformations

mithrandir123:
- Fixed SDF shape transformations (poses defined in model, link, collision, and visual now concatenate in correct order)
 - Changed SDF affordance naming (dropped SDF prefix, zero length names, and collision/visual name suffix when link contains only one element of that type)


patmarion:
Can you upload a video or write a description of how you're using this?  Sounds cool!

mithrandir123:
I'll create a unit test. 
I'm just copy-pasting python commands into the DDApp console at the moment:
```Python
from ddapp import sceneloader
sc=sceneloader.SceneLoader()
sc.loadSDF('absolute_path_tofile.sdf')
```

PR #59, opened 2015-08-19
SDF affordance unit test

mithrandir123:
- Added SDF affordance unit test routine
 - Simplified affordance naming

### DIFF
--- a/src/python/ddapp/sceneloader.py
+++ b/src/python/ddapp/sceneloader.py
@@ -35,7 +35,7 @@ class SceneLoader(object):
                 t.Concatenate(t3)
                 p = transformUtils.poseFromTransform(t)
                 name = model.name;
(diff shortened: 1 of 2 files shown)


patmarion:
Cool, I just loaded tabledemo.sdf in drake-designer.  This is a really nice new feature!

patmarion:
By the way, I see you're creating an individual affordance for each link element.  In the future, would you rather have one affordance for each model element?  I'd like to add a "PrimitiveCollectionAffordanceItem" to support this.  In the meantime, the only option might be the MeshAffordanceItem.

mithrandir123:
The PrimitiveCollectionAffordanceItem sounds good.
I would like to avoid using meshes because I want to work on heavily utilising basic shapes for collision detection in our planners.

PR #65, opened 2015-08-26
DrakeVisualizer: allow different scaling for each dimension

steinachim:
Added the ability to apply different scaling for each dimension (before, only the first scale value was applied for all dimensions)


patmarion:
Thanks!

PR #66, opened 2015-08-30
adding scripts for rrt planning testing

mauricefallon:
These scripts were directly generated from ddapp and test collision free reaching. The key thing is that they use the collisionFreePlanner class and runIKServer - so the tests help maintain integration.

To invoke:
matlab -nosplash -nodesktop -r "addpath(fullfile(getenv('DRC_BASE'),'software','build','matlab')); addpath_control; cd ([getenv('DRC_BASE') '/software/ddapp/src/matlab']); test_rrt_valkyrie() "

@patmarion : I would like to add these tests to the DRC build server. This would require adding
the ctest scripts from software/control to ddapp and making it play well with your python tests.
What do you think about doing this?

### DIFF
--- a/src/matlab/test_rrt_atlas.m
+++ b/src/matlab/test_rrt_atlas.m
@@ -8,8 +8,8 @@
 format long e
 addpath_control
 addpath([getenv('DRC_BASE'), '/software/ddapp/src/matlab'])
-robotURDF = '/home/mfallon/drc/software/models/atlas_v3/./model_convex_hull_robotiq_hands.urdf';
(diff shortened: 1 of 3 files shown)


mauricefallon:
ps: requires fixes to val2 model in drc repo

(atlas and val1 scripts run fine as is)

mauricefallon:
paired with https://github.com/mitdrc/drc/pull/2117

patmarion:
I'll try getting the tests to appear on CDash tomorrow.

PR #68, opened 2015-08-30
multi-seed RRT and RRT* planning

mauricefallon:
Apologies for the triple pull request @patmarion . this is @MarcoCar 's work 

* Demo: https://youtu.be/IfinBB_eudY
* Drake: RobotLocomotion/drake#1302
* mitdrc/drc: https://github.com/mitdrc/drc/pull/2094

These script included, to launch:

matlab -nosplash -nodesktop -r "addpath(fullfile(getenv('DRC_BASE'),'software','build','matlab')); addpath_control; cd ([getenv('DRC_BASE') '/software/ddapp/src/matlab']); test_rrtstar_valkyrie() "


src/python/ddapp/tabledemo.py
```diff
@@ -248,12 +250,8 @@ def segmentTableObjects(self):
     def graspTableObject(self, side='left'):
 
         obj, objFrame = self.getNextTableObject(side)
-        if self.useCollisionEnvironment:
-            objAffordance = om.findObjectByName(obj.getProperty('Name') + ' affordance')
-
+
```
iamwolf: Removing this conflicts with segmenting from live sensor data

src/python/ddapp/tabledemo.py
```diff
@@ -269,17 +267,13 @@ def dropTableObject(self, side='left'):
         self.clusterObjects.remove(obj) # remove from clusterObjects
         om.removeFromObjectModel(obj) # remove from objectModel
```
iamwolf: as above - required for working with real sensor data

src/python/ddapp/tabledemo.py
mauricefallon: @iamwolf : i am aware this probably introduces conflicts in tabledemo.py, I'd like to fix them after the merge as this PR is complicated.

src/python/ddapp/tabledemo.py
```diff
@@ -667,6 +720,9 @@ def playSequenceNominal(self):
         self.planPlaybackFunction(self.plans)
 
     def commitManipPlan(self):
+        if self.planner == 1:
```
iamwolf: Perhaps switch using visOnly or expose visOnly in GUI rather than couple playback/real commit to planner

src/python/ddapp/tabledemo.py
iamwolf: @mauricefallon Okay, I'll have a look at it once I'm back

iamwolf: @mauricefallon can we comment-out rather than delete in this case please?

src/matlab/IKServer.m
```diff
@@ -26,7 +27,7 @@
     function obj = loadNominalData(obj,filename)
       if nargin < 1
         filename = [getenv('DRC_BASE'), ...
-          '/software/control/matlab/data/atlas_bdi_fp.mat'];
+          '/software/control/matlab/data/atlas_v3/atlas_bdi_fp.mat'];
```
patmarion: I'd be in favor of removing this default argument completely.  The caller should always pass in a valid path.

mauricefallon: I agree and obviously v3 hasn't been used in a long time.

Note that I reorganised the folder structure of the fixed points. Would you agree with moving the fixed points out of the control folder and into the specific robot model directories?

patmarion: Yes.  We'd need to update a few locations that assume filename locations in software/control but I think this would be a good improvement, definitely.

And more generally, I'd like to move away from reading nominal pose data from .mat files, and as much as possible pass in the nominal pose as arguments.  We already do this in a lot of places, but there's probably still a few consumers of the .mat files.

src/matlab/IKServer.m
```diff
@@ -174,13 +175,35 @@
...
       options = struct('floating', false);
       xyz = zeros(3,1);
       rpy = zeros(3,1);
       obj.robot = obj.robot.addRobotFromURDF(filename , xyz, rpy, options);
       obj.robot = compile(obj.robot);
     end
+    
+    function obj = addAffordanceToHand(obj, side, affordance, q, name)
+      hand.left = obj.robot.findLinkId('LeftPalm');
+      hand.right = obj.robot.findLinkId('RightPalm');
```
patmarion: I think this looks like a general function, except for these hard coded link names.

src/matlab/IKServer.m
```diff
@@ -174,13 +175,35 @@
...
+      obj.robot.body(hand.(side)).collision_geometry
+      obj.robot = addGeometryToBody(obj.robot, hand.(side), affordance, name);
+      obj.robot = obj.robot.compile();
+      obj.robot.body(hand.(side)).visual_geometry
+      obj.robot.body(hand.(side)).collision_geometry
+    end
+    
+    function obj = removeAffordanceFromHand(obj, side, name)      
+      hand.left = obj.robot.findLinkId('LeftPalm');
+      hand.right = obj.robot.findLinkId('RightPalm');
```
patmarion: likewise

src/matlab/runRRTIKServer.m
```diff
@@ -0,0 +1,38 @@
+s = IKServer();
```
patmarion: Is this file, runRRTIKServer.m called from someplace else?  I don't see any references to it in this pull request.

src/python/ddapp/ik.py
```diff
@@ -369,3 +369,61 @@ def tick(self):
 
         if self.handleAsyncTasks() > 0:
             return
+            
+    def runMultiRRT(self, grHand, qStart, xGoal, objectGrasped):
```
patmarion: this this function be named to include the word "test" since it seems like it's not a generally usable function but just a valkyrie test entry point?

src/python/ddapp/ik.py
```diff
@@ -369,3 +369,61 @@ def tick(self):
...
+        
+        commands = ['info = test_multiRRT_reach(r, s, \'%s\', %s, %s, %d);\n'%(grHand, qStart, xGoal, objectGrasped)]
+        self.comm.sendCommands(commands)
+
+        info = self.comm.getFloatArray('info')[0]
+        
+        return info
+        
+    def addAffordanceToHand(self, handLink, affordance, q, name):
```
patmarion: the arg handLink is actually a side string like "left" or "right"?  At least that appears to be what the matlab function is using, and the actually hand link names are hardcoded.

src/python/ddapp/ik.py
```diff
@@ -369,3 +369,61 @@ def tick(self):
...
+        affStr = ''
+        desc = affordance.getDescription()
+        if desc['classname'] == 'BoxAffordanceItem':
+            affStr += 'RigidBodyBox(%s, %s, quat2rpy(%s));,'%(desc['Dimensions'], list(desc['pose'][0]), list(desc['pose'][1]))
+        elif desc['classname'] == 'CylinderAffordanceItem':
```
patmarion: this function should probably raise an exception for unsupported affordance types

src/python/ddapp/ik.py
```diff
@@ -369,3 +369,61 @@ def tick(self):
...
+        commands.append('aff = %s;\n' % affStr )          
+        commands.append('s = s.addAffordanceToHand(\'%s\', aff, %s, \'%s\');' % (handLink, q, name))
+        self.comm.sendCommands(commands)
+    
+    def removeAffordanceFromHand(self, handLink, name):
+        commands = []        
+        commands.append('s = s.removeAffordanceFromHand(\'%s\', \'%s\');' % (handLink, name))
+        self.comm.sendCommands(commands)
+    
+    def planNominalPose(self, grHand, qStart, stanceFrame):
```
patmarion: I'm not clear what this is doing.  Doesn't look like it belongs here.

src/python/ddapp/ikplanner.py
```diff
@@ -1352,6 +1352,15 @@ def computePostureCost(self, pose):
         return np.sum(np.abs(pose - nominalPose)*cost)
 
 
+    def runMultiRRT(self, grHand, qStart, xGoal, objectGrasped = False):
```
patmarion: internally, it seems this function isn't genreally usable, so it probably shouldn't be here.  Perhaps it should be part of a specific task planner class like the tabledemo.py

mauricefallon:
Thanks for the feedback on this @patmarion , I aim to work on these fixes when the related drake PR has been accepted. Some of the corrections needed here are obvious.

src/matlab/IKServer.m
mauricefallon: Removed hard coded link names

src/matlab/IKServer.m
mauricefallon: Removed hard coded link names

src/matlab/runRRTIKServer.m
mauricefallon: remove it, was being called from mitdrc/drc itself

src/python/ddapp/ik.py
mauricefallon: runMultiRRT() is currently the entry point for this new capability. It has two steps:
* planner.findFinalPose() - find a collision free end pose which can be used instead of runIK
* planner.findCollisionFreeTraj(xGoalFull) - which is the main rrt planner

As you can see it now mirrors runIkTraj(). On another PR I will merge it into runIkTraj().

src/python/ddapp/ik.py
mauricefallon: fixed and generalized

src/python/ddapp/ik.py
mauricefallon: added

src/python/ddapp/ik.py
mauricefallon: removed for now

src/python/ddapp/ikplanner.py
mauricefallon: It cascades om way you have ikPlanner.runIkTraj() call ik.runIkTraj().
Again, I would like to merge runMultiRRT() into runIjTraj() and remove it from here, later if thats ok.

mauricefallon:
Thanks for you patience @patmarion : I'm waiting on the related drake PR, but I think I've improved this code and am done with it.

patmarion:
Ok great, I'll review.

A heads up: commits to drake RL/master within the past weeks have broken compatibility with DRC.  So we'll have to fix DRC again.  Mostly trivial updates to accommodate drake API changes, but it could be a lot of files.  The two changes I know about are changes to the coordinate frame api and the kinematics api, but there may be others.

mauricefallon:
Closing PR, and reopening new version

PR #75, opened 2015-09-17
Updates to tabledemo and shared autonomy

iamwolf:
- **Added shortcut F10 to open the task panel**, similar to F8, F9, and Ctrl+B - since we're finding ourselves opening the task panel a lot. Happy to change to another key, but F10 came in handy after F8, F9
- **Methods to jump in the task tree and continue** - in tabledemo used to repeat the spiel until there are no more objects that have been segmented (clusterObjects is empty). If there's a better way than pausing and continuing, please let me know :)
- **Fix for EXOTica planning when checking manip plan info**: EXOTica mirrors the 0 plan info for successful requests and 13 when constraints are not satisfied, but we also have negative return codes (this shouldn't conflict with Matlab planning - does it?)
- Minor changes that are flagged with the fixed base switch

### DIFF
--- a/src/python/ddapp/tabledemo.py
+++ b/src/python/ddapp/tabledemo.py
@@ -71,6 +71,7 @@ class TableDemo(object):
         self.requiredUserPromptEnabled = True
 
         self.plans = []
+        self.clusterObjects = []
         self.frameSyncs = {}
 
         self.graspingHand = 'left' # left, right, both
--- a/src/python/ddapp/tabledemo.py
+++ b/src/python/ddapp/tabledemo.py
@@ -1300,3 +1300,7 @@ class TableTaskPanel(TaskUserPanel):
             addManipulation(functools.partial(v.planDropPostureLower, v.graspingHand), name='drop: lower arm')
         else:
             addManipulation(functools.partial(v.planPreGrasp, v.graspingHand), name='go home')
+
+        if len(v.clusterObjects) > 0: # There is still sth on the table, let's do it again!
+            self.taskTree.selectTaskByName('reach')
+            self.onContinue()
--- a/src/python/ddapp/tabledemo.py
+++ b/src/python/ddapp/tabledemo.py
@@ -1206,7 +1206,7 @@ class TableTaskPanel(TaskUserPanel):
             if confirm:
                 addTask(rt.UserPromptTask(name='Confirm execution has finished', message='Continue when plan finishes.'), parent=parent)
 
-        def addManipulation(func, name, parent=None, confirm=True):
+        def addManipulation(func, name, parent=None, confirm=False):
             group = self.taskTree.addGroup(name, parent=parent)
             addFunc(func, name='plan motion', parent=group)
             addTask(rt.CheckPlanInfo(name='check manip plan info'), parent=group)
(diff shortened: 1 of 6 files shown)


iamwolf:
@patmarion @mauricefallon are these changes good to be merged?

src/python/ddapp/tasks/robottasks.py
```diff
@@ -178,7 +178,7 @@ def getDefaultProperties(properties):
         properties.setProperty('Message', 'Plan is invalid. Do you want to accept it anyway?')
 
     def run(self):
-        if robotSystem.ikPlanner.lastManipPlan and max(robotSystem.ikPlanner.lastManipPlan.plan_info) <= 10:
+        if robotSystem.ikPlanner.lastManipPlan and max(robotSystem.ikPlanner.lastManipPlan.plan_info) <= 10 and min(robotSystem.ikPlanner.lastManipPlan.plan_info) >= 0:
```
patmarion: I'm not familiar with this, is this saying that sometimes an infeasible plan_info is < 0?

src/python/ddapp/tabledemo.py
```diff
@@ -1194,6 +1196,14 @@ def onPropertyChanged(self, propertySet, propertyName):
         elif propertyName == 'Scene':
             self.tableDemo.sceneID = self.params.getProperty('Scene')
 
+    def pickupMoreObjects(self):
+        if len(self.tableDemo.clusterObjects) > 0: # There is still sth on the table, let's do it again!
+            print "There are more objects on the table, going at it again!"
+            self.taskTree.selectTaskByName('reach')
```
patmarion: I think this is probably working very well for you.  It's also possible to make a task that adds new tasks to the queue.  The result would be that list of tasks grows each loop, instead of returning to the start, which may or may not be desirable.

src/python/ddapp/tasks/robottasks.py
iamwolf: Yes, exactly -- is this a problem for drake or the current system?

patmarion:
+1 looks good.  Only thing I'd be worried about is the new check for negative plan_info.  The problem is that there are a few other places in the code that check plan_info, and have hard coded assumptions about what is/isn't feasible.  For example, in playbackpanel.py the functions
isPlanFeasible() and getPlanInfo().  Maybe we could move the isPlanFeasible() check to a single reusable utility function in a common module.

src/python/ddapp/tabledemo.py
patmarion: I think your looping strategy is good.  The only issue is that it has to know the name of the starting task, 'reach' in this case, but I think that's fine.

iamwolf:
@patmarion Thanks for the feedback. I've checked and the same assumption is used in playbackpanel as you mentioned as well as in valvedemo. I've changed it in both places as well and added it to this PR. I think moving to a global function in the mid-term is a good idea. (happy to add it to my todo list; we can also write those assumptions down somewhere in a wiki or so in case somebody else wants to implement or expand on external planning)

patmarion:
Thanks!

patmarion:
oops, looks like a small typo:

  File "/home/pat/source/drc/drc/software/build/lib/python2.7/dist-packages/ddapp/playbackpanel.py", line 261
    return plan is not None and max(plan.plan_info) < 10 and min(plan.plan_info) >= 0:

iamwolf:
Maurice just caught that as well and will submit the fix with #78 - apologies for overlooking this

PR #77, opened 2015-09-27
use new lcmtypes package

rdeits:
to support https://github.com/RobotLocomotion/drake/pull/1380


mauricefallon:
If the director submodule is referenced in drake is this necessary?

patmarion:
@mauricefallon are you asking if the pull request is necessary?  As in, why merge, instead of just referencing the director branch from drake?  I think this pull request is ok, because it has been written to be backward compatible, and will be necessary when DRC eventually updates its own drake submodule reference.

patmarion:
So I think this one is good, +1.  I'll can for more comments and merge a later today.

mauricefallon:
ok, i see, fine by me

PR #81, opened 2015-10-01
Use hand grasping status in tabledemo (no more human acknowledgement required)

iamwolf:
This series of commits allows to run the tabledemo even more autonomous and recognise whether the hand has grasped something - i.e. it does not require human verification of that the hand is closed and in fact has an item within its fingers.

- Added a new status topic ``GRASPING_STATE`` (boolean_t) that records whether an item has been grasped (``true``) or the hand was successfully opened (``false``)
- Implemented the above as a grasping task in robottasks -- waits (blocking) until a message is received or fails with a timeout (and then blocks further execution of the task tree)
- Added an addGrasping function for this
- Deactivated for humanoids and all others that dont have this message implemented yet
- Changed a few flags to check whether we are dealing with kukas rather than with a standard fixed base

### DIFF
--- a/src/python/ddapp/tasks/robottasks.py
+++ b/src/python/ddapp/tasks/robottasks.py
@@ -711,7 +711,7 @@ class CloseHand(AsyncTask):
         if self.properties.getProperty('Check status'):
             responseChannel = 'GRASPING_STATE'
             responseMessageClass = lcmdrc.boolean_t
-            grasping_state = lcmUtils.MessageResponseHelper(responseChannel, responseMessageClass).waitForResponse(timeout=5000)
+            grasping_state = lcmUtils.MessageResponseHelper(responseChannel, responseMessageClass).waitForResponse(timeout=7000)
             if grasping_state is not None and grasping_state.data == 1:
                 self.statusMessage = "Grasping successful"
             else:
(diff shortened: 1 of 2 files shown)


src/python/ddapp/tasks/robottasks.py
```diff
@@ -706,6 +708,15 @@ def run(self):
...
 
+        if self.properties.getProperty('Check status'):
+            responseChannel = 'GRASPING_STATE'
+            responseMessageClass = lcmdrc.boolean_t
+            grasping_state = lcmUtils.MessageResponseHelper(responseChannel, responseMessageClass).waitForResponse(timeout=7000)
+            if grasping_state is not None and grasping_state.data == 1:
+                self.statusMessage = "Grasping successful"
+            else:
+                self.fail("No object in hand")
+
```
patmarion: You might consider putting this code into a new task called WaitForGraspingState or something like that.  Then you can replace these lines with:

   WaitForGraspingState().run()

That will be useful in case people want to parameterize the WaitForGraspingState task in the future.

iamwolf: Great idea! Would this have the WaitForGraspingState task as a subtask of this one? I.e. so one property in the Grasping task can determine whether to wait or not?

patmarion: You could keep the 'Check status' property here.  The value of 'Check status' determines whether or not WaitForGraspingState is invoked here.  So the behavior would be exactly the same.

Or, if you wanted, you could remove the 'Check status' property, and instead the task queue could have a CloseHand task followed by a WaitForGraspingState task.  In some ways, this strategy might be more clear.  Because the CloseHand task succeeds, but  the failure might be during WaitForGraspingState.  Of course, then 'resuming' the task queue would resume at the WaitForGraspingState, when you might really want to resume at CloseHand (to retry).  So that argues for keeping CloseHand and Wait a single task, as you currently have it.

patmarion:
Did you close this intentionally?

iamwolf:
Yes, I am working on changing it to use WaitForGraspingState but couldn't complete testing and committing all changes yet as I got busy with other stuff. Will reopen as soon as it's done and working :)

iamwolf:
@patmarion I've changed it to WaitForGraspingState and tested on hardware (it works)

PR #82, opened 2015-10-05
Continuous walking update (working)

simalpha:
Continuous walking update (working). Task panel modified.


mauricefallon:
This looks fine. In general, I'd suggest deleting unused code instead of commenting things out.

PR #86, opened 2015-10-10
Added contact groups and test function.

simalpha:
Added contact groups and test function for continuous walking.


patmarion:
+1 from me.  Ok to merge?

In the future we need to remove hard coded contact points from footstepsdriver.py, but this change is fine given that we already had all those numbers.

Please write python if statements without parenthesis around the condition, they are not needed.  As in:

    if robotType == 0:

instead of

    if (robotType == 0):

mauricefallon:
I'm happy to merge it too.
I'll put it on my list to remove the hard coded contact points and query the urdf directly..

patmarion:
A while back I added API to query the contact points.  I need to review the reason, but I think I determined that the retrieved contact pts were not symmetric between left and right foot.  At the time, there wasn't code to distinguish between left foot and right foot contact pts (it was assumed they were symmetric w.r.t to each foot frame) so I just left the hard coded points.  But I think now we distinguish between left and right (as it was required for Valkyrie) so we can probably update the code now to do the right thing.  There might also be some special handling required to retrieve contact pts only for just toe/heel and not the midfoot.

PR #88, opened 2015-10-12
clean up the output of the constraintencoder

mauricefallon:
previously the output string encoding a numpy array was as follows:
     "jointsUpperBound": {"__ndarray__": [0.0, 0.0, 0.0]},

@patmarion : is this a robust away of spotting an np.array and converting it to a list?

### DIFF
--- a/src/python/ddapp/ikconstraintencoder.py
+++ b/src/python/ddapp/ikconstraintencoder.py
@@ -11,7 +11,7 @@ import drc as lcmdrc
 import pprint
 import json
 
-class ConstraintEncoder(nje.NumpyEncoder):
+class ConstraintEncoder(nje.NumpyConvertEncoder):
     def default(self, obj):
         if isinstance(obj, vtk.vtkTransform):
             pos, quat = transformUtils.poseFromTransform(obj)
@@ -22,8 +22,7 @@ class ConstraintEncoder(nje.NumpyEncoder):
             for key in obj._fields:
                 d[key] = getattr(obj, key)
             return d
-        return nje.NumpyEncoder.default(self, obj)
-
+        return nje.NumpyConvertEncoder.default(self, obj)
 
 
 def ConstraintDecoder(dct):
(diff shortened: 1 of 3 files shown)


patmarion:
We could add an option to tell the encoder to convert small ndarray objects into lists.  This would be a lossy encoder, because it tosses out information about the original object type, but for the constraint encoder that might be acceptable and preferred.  The file to modify is numpyjsonencoder.py and this is the line that could be modified:

    return dict(__ndarray__=obj.tolist())

change to

    return obj.tolist()

src/python/ddapp/ikconstraintencoder.py
```diff
@@ -15,14 +15,16 @@ class ConstraintEncoder(nje.NumpyEncoder):
...
         elif isinstance(obj, fieldcontainer.FieldContainer):
             d = OrderedDict()
             d['class'] = type(obj).__name__
             for key in obj._fields:
                 d[key] = getattr(obj, key)
             return d
-        return nje.NumpyEncoder.default(self, obj)
+        # return nje.NumpyEncoder.default(self, obj)
+        # Return Numpy array as list
+        return obj.tolist()
```
patmarion: do we want to make this change?  It assumes the object is a numpy data type.  Maybe in this context that is a correct assumption, not sure.

mauricefallon:
We are parsing the json constraints output by director in C++, but for ndarrays, it inserts info about the python type into the output sting.
We aren't concerned with retaining the ndarray information.

I made your suggested change. Does it look ok? I don't know how widely this encoder is used.

patmarion:
Let me take a pass over this.  I'm just going to add an option the this class so that by default it keeps the information, but the constraint encoder can run it in a mode that makes simplified output when it is known that the decoder doesn't use numpy.

patmarion:
all set.  sorry, I ended up overwriting your commits :(

mauricefallon:
no worries, thanks to looking into this!

PR #90, opened 2015-10-16
multi-seed RRT and RRT* planning

mauricefallon:
Adds RRT* planning and final/end pose search:
https://www.youtube.com/watch?v=ww2q2PvQ_mw

This reopens a month old pull request (https://github.com/RobotLocomotion/director/pull/68)
which was reviewed back when updating drake was blocking things, but now this is resolved.


patmarion:
I just read through the changes here.  Looks pretty good!  I will test it on my computer and let you know.  There are a few things I would like to change, but I think we can move ahead with the merge and open new issues to do the changes in the future.  But let me test before merge.

patmarion:
I had some time to test this branch this morning.  I'm mostly fine with merging this, even though I'm not sure how to test all the new features, I was able to test most features and everything still seems to be working fine, so that's good :)  However, one bug: this branch has changed the ikParameters.useCollision variable from a boolean to a string.  But, not all the code that uses the variable has been updated.  For example, in ik.py and teleoppanel.py, there are still a few places testing:  "if params.useCollision", well now that it is a string, this test is always true.  The result is that collision planning is always enabled.  Instead, it should best testing the string value of useCollision.  But, maybe useCollision should become an integer variable with constants defined, so you could test like:

if params.useCollision == params.NO_COLLISION

something like that?

mauricefallon:
Ok, thanks for the suggestion. I agree - using a constant across the board would be preferred.

I can I follow this up after the branch is integrated?

patmarion:
Not quite yet.  It doesn't have to be the full change I suggested (changing the variable from a string to an int) but the if conditions in ik.py and teleoppanel.py need to be updated.  Basically, "grep useCollision" and update the if conditionals that are testing "if useCollision" to "if useCollision == 'none'"

mauricefallon:
Ok, I made those changes and it looks good with a little testing.

PR #92, opened 2015-10-19
 Updated feet contact points position according to SCS code.

simalpha:


mauricefallon:
@simalpha : the y-values aren't updated here. can you find any information about the foot width in SCS?

PR #96, opened 2015-10-23
Reduce required config fields

patmarion:


patmarion:
This branch is to better support arm manipulators.  It removes some of the assumptions about specialized fields that were added to support atlas and drc.  A larger overhaul is definitely needed, but this is just some small stuff for now.

iamwolf:
:+1:

PR #100, opened 2015-10-28
Added shortcuts file to give gentle introduction/list/doc about which…

iamwolf:
… shortcuts exists to onboard new users more easily


patmarion:
Hmm, maybe instead of adding a top level file, we could add this to the docs folder?
maybe something like: docs/apps/drake-designer/shortcuts.md

I would really like to have a Help icon on the toolbar that would launch an in-application doc window with this type information (and lots more!)

iamwolf:
Very happy with the location you think is best to place this file. My rationale for the top level was that this is where people look first, next to the README. Few people go into docs, make the docs, or even go to a GitHub wiki unless pointed to. Happy to add it to the README as well; not sure whether reStructured text supports <kbd>keyboard</kbd> markup.
I do agree, a help icon would likely be the best but also requires more work to add.

patmarion:
I agree that people won't find easily find the docs on the filesystem.  But still, I'd prefer not to add this file at the top level.  We need to have an established location for drake-designer user documentation.  The top level README could point to that spot.

I think an in-application help button could be the obvious place to launch the docs.  It wouldn't be that much work.  As a first pass, the button could just launch a url using the default web browser:

QtGui.QDesktopServices.openUrl(QtCore.QUrl('http://www.github.com/RobotLocomotion/director.git'))

iamwolf:
@patmarion thanks for the feedback. I've moved it to a new subfolder (``docs/drake-designer``), added a main README.md and linked to the SHORTCUTS.md. Also added a Help toolbar menu with Online Help as an action linking to the online help as suggested (the link will only work once this PR is merged). 

I am very happy to add two new shortcuts within the same PR to address https://github.com/openhumanoids/main-distro/issues/1848 -- e.g. <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd> for planning in the teleop panel (I think Ctrl+Enter already does execution?). What do you think? Should we add this or just close https://github.com/openhumanoids/main-distro/issues/1848

patmarion:
Yeah, I think that sounds like a good idea, please add it!

iamwolf:
Added, ready to merge on my part :) 

Closes openhumanoids/main-distro#1848

PR #102, opened 2015-10-30
Fix drc-testing-data filename

iamwolf:
@mauricefallon please sign off on this and confirm that this change is okay. We didn't see the ...-new.vtp and since it is working with the ihmc.vtp I changed the filename. If it is not the right move, please add the -new.vtp file to drc-testing-data -- thanks!


mauricefallon:
i don't see the -new file on machine. yes, lets go with what you are suggesting there

patmarion:
Ok ready to merge then?

PR #104, opened 2015-11-02
Fix and format valvedemo

iamwolf:
It works sort of with IHMC's SCS. There are a few outstanding issues, e.g. lots of infeasible plans because SCS goes beyond joint limits (issue reported) which kills planning and leads to infeasible constraints (the neck is wobbling as a result for example).

Issue List ValveDemo:
- sometimes 'wait for walking' throws an exception because it is already executing -- i.e. our state publisher isn't always working as intended (on the other hand, sometimes it works -- appears to be a timing thing)
- anklePitches over jointLimit (same as neckPitches -- reported to IHMC Bitbucket Issue tracker)
- the plans for the actual turning motion are invalid since it would kill its neck

It's currently in conflict but I'll fix that in due course and comment once it's ready to merge

### DIFF
--- a/src/python/ddapp/valvedemo.py
+++ b/src/python/ddapp/valvedemo.py
@@ -37,7 +37,7 @@ class ValvePlannerDemo(object):
         self.rhandDriver = rhandDriver
         self.sensorJointController = sensorJointController
         self.graspingObject = 'valve'
-        self.turningMode = 0 # Simple Mode
+        self.turningMode = 0  # Simple Mode
         self.setGraspingHand('left')
         self.graspFrame = None
         self.stanceFrame = None
@@ -74,8 +74,6 @@ class ValvePlannerDemo(object):
 
         self.nominalPoseName = 'q_valve_nom'
         self.startPoseName = 'q_valve_start'
-
-
 
     def setGraspingHand(self, side):
         self.graspingHand = side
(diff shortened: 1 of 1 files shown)


iamwolf:
Resolved the merge conflict, so ready to merge from my side :)

src/python/ddapp/valvedemo.py
```diff
@@ -1195,7 +1196,7 @@ def addManipulation(func, name, parent=None):
         addTask(rt.PlanPostureGoal(name='plan arm up pregrasp', postureGroup='General',
                                    postureName='arm up pregrasp', side=side), parent=preturn)
         addTask(rt.CommitManipulationPlan(name='execute manip plan',
-                                          planName='arm up pregrasp plan'), parent=preturn)
+                                          planName='arm up pregrasp posture plan'), parent=preturn)
```
iamwolf: This is the real change in this PR - the rest is autoformatting. Key to note that PlanPostureGoal plans are now '...posture plan' instead of '... plan'. The other issues we are seeing with valveDemo are SCS based (doesnt adhere to joint limits, does not correctly execute trajectories)

patmarion:
By the way, I like having good formatted code.  But I also prefer to not touch lots of lines just to improve formatting (ideally the code would be pep8 compliant in the first commit).  So maybe don't continue running existing files through pep8ify.  In theory we could run a code format checker as a test and require that pull requests conform, but I worry that would slow down development too much.

PR #106, opened 2015-11-03
Make Matlab commands log file user specific

iamwolf:
Changes the path from ``/tmp/matlab_commands.m`` to ``~/.drake-designer/matlab_commands.m``. It is hence user-specific and there won't be conflicts if drake-designer is used in multiple user accounts on the same machine without restarting.

Addresses https://github.com/openhumanoids/main-distro/issues/2177

### DIFF
--- a/src/python/ddapp/matlab.py
+++ b/src/python/ddapp/matlab.py
@@ -147,8 +147,8 @@ class MatlabCommunicator(object):
 
     def getLogFile(self):
         if self.logFile is None:
-            if not os.path.exists(os.path.expanduser('~/.drake-designer')):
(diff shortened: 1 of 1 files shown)


src/python/ddapp/matlab.py
```diff
@@ -147,6 +147,9 @@ def checkForResult(self):
 
     def getLogFile(self):
         if self.logFile is None:
+            if not os.path.exists(os.path.expanduser('~/.drake-designer')):
```
patmarion: I'd change this to os.path.dirname(self.logFileName)

iamwolf:
Thanks for the feedback, I've amended it @patmarion

PR #109, opened 2015-11-09
Strategy for climbing down staircase.

simalpha:
Strategy for climbing down staircase.


src/python/ddapp/terrain.py
```diff
@@ -10,11 +10,13 @@
 from polyhedron import Vrep, Hrep
 from py_drake_utils.utils import rpy2rotmat
 
+FOOTSIZE_REDUCTION = 0.04
+FOOT_LENGTH = 0.25 - FOOTSIZE_REDUCTION
+FOOT_WIDTH = 0.15 - FOOTSIZE_REDUCTION
 
+DEFAULT_FOOT_CONTACTS = np.array([[-0.5*FOOT_LENGTH, -0.5*FOOT_LENGTH, 0.5*FOOT_LENGTH, 0.5*FOOT_LENGTH],
+                                  [0.5*FOOT_WIDTH, -0.5*FOOT_WIDTH, 0.5*FOOT_WIDTH, -0.5*FOOT_WIDTH]])
```
patmarion: Looks like this is changing some values?  I guess this only affects continuous walking and not used elsewhere?  We should try to remove the magic numbers from this file at some point.

Issue #111, opened 2015-11-10
Support URDF colors


patmarion:
This has been a long standing issue.  By default, textures are enabled and URDF colors are disabled.  Things to do:

 - add new color ivar for each link in ddDrakeModel.cpp to store URDF color specified in XML
 - on python RobotModelItem remove "Textures" boolean property, instead we should have a "Color Mode" with three enum options: Solid Color, URDF Colors, Texture
 - add config parameter to director_config.json so that a robot model can specify its preferred default

@iamwolf you can enable URDF colors by modifying roboturdf.py:

set this to True on this line:

        self.useUrdfColors = False

set this to False on this line:

    self.addProperty('Textures', True)

The problem is, if Textures are enabled then all the link colors get set to white.  Since ddDrakeModel.cpp doesn't store URDF colors, once you set the link colors to white there is no way to set them back to urdf colors.  So URDF colors only work if you set the Textures property to default to False.

patmarion:
I'd like to keep this open until the Color Mode property is added, I'll take the assignment.

PR #113, opened 2015-11-11
Fix for #111 - make URDF colors usable again

iamwolf:
This is a working fix for #111, tested with Atlas v3, v4, v5, Valkyrie v1, v2, and the Kuka LWR. It sets the rendering mode to URDF Colors by default and overwrites it to "Texture" if given in the directorConfig

Goes along with https://github.com/openhumanoids/main-distro/pull/2196

### DIFF
--- a/src/python/ddapp/roboturdf.py
+++ b/src/python/ddapp/roboturdf.py
@@ -186,7 +186,7 @@ class RobotModelItem(om.ObjectModelItem):
         view.render()
 
 
-def loadRobotModel(name, view=None, parent='planning', urdfFile=None, color=None, visible=True, colorMode='URDF Colors'):
+def loadRobotModel(name, view=None, parent='planning', urdfFile=None, color=None, visible=True, colorMode='Solid Color'):
 
     if not urdfFile:
         urdfFile = urdfConfig['default']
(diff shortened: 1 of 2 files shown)


iamwolf:
Note, this does not read the URDF Colors into the ddDrakeModel so hot-swapping rendering modes during runtime won't work, but that's not a use-case that we are pursuing

src/python/ddapp/roboturdf.py
```diff
@@ -203,6 +204,11 @@ def loadRobotModel(name, view=None, parent='planning', urdfFile=None, color=None
     obj.setProperty('Visible', visible)
     obj.setProperty('Name', name)
     obj.setProperty('Color', color or getRobotGrayColor())
+    if colorMode == 'Texture': # fix for #111
+        obj.setProperty('Textures', True)
+    elif colorMode == 'URDF Colors':
```
patmarion: since useUrdfColors is already set to True in the RobotModelItem constructor, shouldn't this be handling the Solid Color case? Set useUrdfColors to false and call obj._updateModelColor()

src/python/ddapp/roboturdf.py
```diff
@@ -185,7 +186,7 @@ def removeFromView(self, view):
         view.render()
 
 
-def loadRobotModel(name, view=None, parent='planning', urdfFile=None, color=None, visible=True):
+def loadRobotModel(name, view=None, parent='planning', urdfFile=None, color=None, visible=True, colorMode='URDF Colors'):
```
patmarion: I don't think colorMode should default to URDF Colors until the option is switchable by the user.

patmarion:
Thanks for working on this.  It would be really nice to fix this completely :)  But I'm fine to merge the partial fix since it adds the config option.

iamwolf:
Thanks for the comments and feedback, I've amended the changes accordingly. I think for now this is a good partial fix; we can always go back and change the ddDrakeModel properly when needed :)

PR #114, opened 2015-11-11
URDF/Texture Solid Color Toggling Fix

iamwolf:
This fixes a bug @mauricefallon spotted that I introduced in #113 - toggling solid colors was broken when operating with textures. This PR allows for toggling between solid colors and textures, e.g. for the playback model


patmarion:
Thanks!  When we eventually add the functionality to ddDrakeModel.cpp, none of the special logic in the Python layer will be required, because the ddDrakeModel object will know how to properly switch between all the modes.

PR #115, opened 2015-11-12
allow runtime config designer

mauricefallon:
this evaluates a list of strings at the top of startup.py and disables the components e.g.:
this disables drake planning but will draw a robot and render sensor data:
  "disableComponents" : ["useFootsteps","usePlanning","useFootContactVis","useCOPMonitor"]

### DIFF
--- a/src/python/ddapp/startup.py
+++ b/src/python/ddapp/startup.py
@@ -171,7 +171,7 @@ if 'fixedBaseArm' in drcargs.getDirectorConfig()['userConfig']:
 if 'disableComponents' in drcargs.getDirectorConfig():
     for component in drcargs.getDirectorConfig()['disableComponents']:
         print "Disabling", component
-        exec(component + " = False")
+        locals()[component] = False
 
 
 if useSpreadsheet:
--- a/src/python/ddapp/startup.py
+++ b/src/python/ddapp/startup.py
@@ -166,17 +166,12 @@ useMappingPanel = True
 poseCollection = PythonQt.dd.ddSignalMap()
 costCollection = PythonQt.dd.ddSignalMap()
 
-
 if 'fixedBaseArm' in drcargs.getDirectorConfig()['userConfig']:
     ikPlanner.fixedBaseArm = True
-if 'useFootstepsDisable' in drcargs.getDirectorConfig()['userConfig']:
-    useFootsteps = False
-if 'usePlanningDisable' in drcargs.getDirectorConfig()['userConfig']:
-    usePlanning = False
-if 'useFootContactVisDisable' in drcargs.getDirectorConfig()['userConfig']:
-    useFootContactVis = False
-if 'useCOPMonitorDisable' in drcargs.getDirectorConfig()['userConfig']:
-    useCOPMonitor = False
+if 'disableComponents' in drcargs.getDirectorConfig():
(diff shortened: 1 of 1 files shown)


mauricefallon:
For example see this director config file, with the disableComponents field

https://github.com/drcbot/main-distro/blob/4b0a5267712875a68e12de50737cea83880892fb/software/models/common_components/multisense_sl/director_config.json

patmarion:
Cool!  First of all, I'm happy that the application actually starts up and is functional with footsteps and planning disabled, that's nice!

I put all the useXX boolean variables in startup.py because I knew one day we'd want to toggle features on off, just like you used to do with the renderers of the bot viewer.  But of course, we never test whether turning off a component causes runtime errors because of componenter inter-dependencies.

I'm happy to merge this now, because it's already quite useful.

But I'm hoping we can take some time (maybe tomorrow's phone call) to discuss the full implications of this change.  In fact, we've never really had a deep discussion about where we want to go with director_config.json in general.  This pull request adds a new key named disableComponents.  It works, but I'm sure we could invest some time designing a mechanism that would be flexible and scale into the future. I want to be careful not to continue patching the json file without some long term vision.

src/python/ddapp/startup.py
```diff
@@ -166,6 +166,14 @@
 poseCollection = PythonQt.dd.ddSignalMap()
 costCollection = PythonQt.dd.ddSignalMap()
 
+if 'fixedBaseArm' in drcargs.getDirectorConfig()['userConfig']:
+    ikPlanner.fixedBaseArm = True
+if 'disableComponents' in drcargs.getDirectorConfig():
+    for component in drcargs.getDirectorConfig()['disableComponents']:
+        print "Disabling", component
+        exec(component + " = False")
```
patmarion: slightly better:

locals()[component] = False

But in the follow up commit we could introduct an Options struct instead of having the useXX global variables.
For example:

from ddapp.fieldcontainer import FieldContainer

    opts = FieldContainer(
      useFootsteps = True,
      usePlanning = True
    )

    # now we can access opts.usePlanning

mauricefallon:
I tested and pushed a commit to use 
locals()[component] = False

I think you were implying the fieldcontainer would be used in future and not for this PR?

mauricefallon:
Like you say toggling on and off components is not straightforward as several components are interdependent. Judging from this exploration there may be 2-3 groups of components that are independent currently.

But I agree, the JSON is used as a stopgap at present

patmarion:
Yeah I think in some future work we could make this a stronger feature.  It's essentially a plug-in system, where components can be added/removed to a general purpose base application.  We'll need need a mechanism that can express interdependencies possibly, and definitely a testing framework to ensure the various combinations continue to work together correctly.

So far, we've gotten away without extensive test suite because there is mostly just one operation mode so we developers are quickly aware if something breaks.  Now to support many optional components, we need much more test coverage.  For starters, I'd like to introduce some tests that demonstrate how to launch drake-designer, execute some code, and exit.

A concern I'd like to mention is that this mechanism can allow us to add new components that are disabled by default.  The component might only be relevant to one person's use case.  These types of components can become a maintenance burden.  Since they are off by default, we just say yeah sure go ahead and merge it, even though the code might not otherwise be up to our standard.

PR #118, opened 2015-11-18
rename ddapp and designer to be director

mauricefallon:
I pretty much did exactly as @patmarion suggested:
software/ddapp --> software/director
src/python/ddapp --> src/python/director
"import ddapp" --> "import director"
drake-designer --> director
ddConsoleApp --> directorPython
c++ classes can keep the "dd" prefix for now, as in ddRobotModel.cpp

Tested with simulation (val and atlas), some logs and manually running a unit test.

Probably need to update the dashboard on the build server

The follow on changes needed for drake are in:
drake/examples/Pendulum/Pendulum.pmd
drake/examples/Pendulum/CMakeLists.txt
drake/systems/plants/BotVisualizer.m
... but they are pretty straightforward


patmarion:
I'm excited about this!

PR #119, opened 2015-11-18
Slider board midi

mithrandir123:
- Added support for Behringer BCF2000 midi slider board.
 - Hooked up the slider board to joint teleop panel.

When the board is not connect, joint teleop behaves as usual.

https://youtu.be/Nwju1w1DaPM

### DIFF
--- a/src/python/director/teleoppanel.py
+++ b/src/python/director/teleoppanel.py
@@ -15,6 +15,7 @@ from director.roboturdf import HandFactory
 from director import lcmUtils
 import drc as lcmdrc
 
+import director.midi as midi
 import functools
 import math
 import numpy as np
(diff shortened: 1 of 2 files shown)


src/python/director/teleoppanel.py
```diff
@@ -1107,6 +1154,27 @@ def buildTabWidget(self, jointGroups):
 
         self.signalMapper.connect('mapped(const QString&)', self.sliderChanged)
 
+    def tabChenged(self, id):
```
iamwolf: *chAnged

patmarion:
Hey this is great!  I want to get one of those!

I think this is awesome, but I don't want to merge this change as is.  I think it needs to be designed so that the midi is decoupled from the UI.  The panel class shouldn't have a behringer member variable, and it shouldn't have a member function named setupSliderBoard().  Likewise, a function like slider board that deals with midi values shouldn't know anything about joint names.  Instead, maybe we could design a very general class that can manage the sync between a Qt slider and a midi device.  Next, the teleop panel class can be extended to provide whatever general purpose API is required to attach a specific instance of this manager object which handles the mapping of joint names to sliders.  I know this isn't a very clear description, but something like that.  Ideally, the only changes to teleoppanel.py are just changes that add generic api.

mithrandir123:
I was afraid you were going to say exactly that :)
Alas, I agree.

patmarion:
I'm going to close this for now.  But I hope the work doesn't disappear, I think it's really great, just needs to be moved into a standalone module and then it will be easier to integrate.

mithrandir123:
It's still on my list. I'll reopen the pull request when I get to it.

PR #120, opened 2015-11-19
Documentation

patmarion:


patmarion:
I added a new layout for sphinx documentation.  Sphinx is a tool that generates static html files from restructured text files (.rst file extension).  Most of the rst files I added are stubs, but please have a look and let me know what you think.  Everything is in docs/sphinx.  Here is what the output looks like:

    http://openhumanoids.github.io/director/

Please let me know whether or not you think it's a good idea to use the github.io url.  I put it there for now since it was easy.


If you want to generate the html documentation yourself, it's quite easy once you have sphinx installed.  On ubuntu I installed sphinx with:

    sudo pip install sphinx sphinx-rtd-theme


You can generate sphinx documentation with:

    cd director/docs/sphinx
    make html

    # view the result
    firefox _build/html/index.html


To make sphinx documentation that includes the python module auto documentation, you need to run these two commands:

cd director
make -C distro/pods/drc/pod-build docs-sphinx-generate
make -C distro/pods/drc/pod-build docs-sphinx-build


The docs-sphinx-generate target will generate module rst files in docs/sphinx/generated.  If you run make html then you will get a bunch of import errors because sphinx doesn't know how to import the director modules to generate documentation, that's why you need to run the docs-sphinx-build command as written above.

PR #122, opened 2015-11-19
Add octomap support

mauricefallon:
Based off of the LCMGL vtk renderer and the Octomap libbot renderer, I created this mash up.
The rendering GL code originated in the actual octomap library.
https://www.dropbox.com/s/k73quafcyj60hnl/octomap-in-director.m4v?dl=0
for reference:
https://www.dropbox.com/s/09lnwkeudmydci1/2015-09-fusion01b.mp4?dl=0

This is not fully finished feature, it doesn't have these features from the libbot version:
- the properties are not properly sync'ed between python and c++ yet:
  - no min/max properties for showing part of a map
  - no min/max properties for coloring red/blue 
- also ~2 transmits of the octree are needed to send the cloud - even though the data arrives in the designer. not sure why.

Also includes an libbot function to return a jet color. maybe this is not needed.

@iamwolf  is going to take it on know, but wanted to get feedback from @patmarion at this stage.


patmarion:
Can you share a log with the octomap messages or tell me how to run the process that publishes octomap from lidar?

mauricefallon:
the .bt files here are in octomap format:
https://www.dropbox.com/sh/pmny97snn1tnu55/AACju6ih9R-UIxQNG6dXW7SRa?dl=0

to publish the octomap as an LCM message:
octomap-server octomap.bt

if you want to see a side-by-side, the old gtk viewer renderers in the pronto viewer:
pronto-viewer -c atlas/robot.cfg

patmarion:
Cool, thanks.  I hope to try it tomorrow.  I read thru the code in this pull request, looks great.  Plus, you took care of the all fun stuff adding the octomap options to cmakelists, nice!

I'm not sure how performance will vary (improve or not) but there may be some advantages to taking the approach where you generate geometry and let VTK handle the opengl drawing.  But, this current implementation seems to work fine, too.  In your video the framerate seemed decent.  It looked like there was not lighting / illumination though, just flat color.  Maybe I'm wrong about that.  Perhaps that's easy to toggle on as one of the opengl bits?

Let me know what you think, I'm happy to merge this now if you'd like to get it in, and you guys can do more work on it from there.

mauricefallon:
Rendering performance seems to be fine. (its a function of the level of the tree). It seems to render as efficiently as the other version.

You are right, the rendering is just flat color. I haven't dug into the opengl code to look at illumination and alpha. I'm happy to bring it in and improve it from there.

We are working on connecting collision free planning to real perception data.

mauricefallon:
BTW, more recent version of the rendering library comes with the octomap library itself. It looks like it wouldn't be difficult to move to using it.

This is their own application built on it:
https://www.dropbox.com/s/361fkx1w5uthc34/octovis-gui.m4v?dl=0

patmarion:
+1 for upgrading OctoMap.  I think there was something like a compile error in common_utils, perhaps the libbot renderer or something like that.  But I think it's worth upgrading and fixing the compile error if we can.

patmarion:
Note, here is some discussion about upgrading OctoMap and the associated compile errors:

https://github.com/openhumanoids/main-distro/issues/2067

PR #123, opened 2015-11-19
Plane segmentation filters

patmarion:


patmarion:
Adds VTK filters and a python test.  Here is output of the test script on DRC trials terrain:

plane-segmentation

PR #126, opened 2015-11-23
Color Mode, per #111

### LINKED ISSUE #111: Support URDF colors

iamwolf:
Pat, apologies for closing the issue without adding the color mode enum first (I wasnt aware of that fact and should have asked before closing).

Were you looking for something along the lines of this? This adds a enum for Color Mode and populates it with "Solid Color" per default, and then changes it according to a director config parameter to Textures or URDF Colors. I also changed "Texture" to "Textures" to have a common naming convention, cf. https://github.com/openhumanoids/main-distro/pull/2230.

This removes the Textures boolean property and replaces it with a Color Mode enum. Tested on v5, val2, lwr.

screenshot from 2015-11-23 17 25 28

Addresses, and potentially closes #111

### DIFF
--- a/src/app/ddDrakeModel.cpp
+++ b/src/app/ddDrakeModel.cpp
@@ -76,6 +76,7 @@ class ddMeshVisual
   vtkSmartPointer<vtkTransform> Transform;
   vtkSmartPointer<vtkTransform> VisualToLink;
   vtkSmartPointer<vtkTexture> Texture;
+  QColor Color;
   std::string Name;
 
 private:
@@ -667,6 +668,7 @@ public:
           meshVisual->Name = body->linkname;
           meshMap[body].push_back(meshVisual);
 
+          meshVisual->Color = QColor(visual.getMaterial()[0]*255, visual.getMaterial()[1]*255, visual.getMaterial()[2]*255);
           meshVisual->Actor->GetProperty()->SetColor(visual.getMaterial()[0],
                                                      visual.getMaterial()[1],
                                                      visual.getMaterial()[2]);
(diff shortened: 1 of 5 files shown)


patmarion:
Yes but I think that if you switch from URDF colors to solid color or textures, then won't be able to switch back to URDF colors, because the c++ class ddDrakeModel doesn't actually store the URDF colors using ivars.  Just a little more book keeping required at the c++ layer.  Once the c++ class switches colors correctly, then the weird python logic (self.useUrdfColors) can be removed completely.

iamwolf:
@patmarion Please have a look at the updated code. Is this along the lines you were thinking? Also cf. https://youtu.be/_0msnEZQh6U

patmarion:
Awesome!  Thanks for putting this together, this is really nice to have.

PR #133, opened 2015-11-27
Add mesh scale

patmarion:


patmarion:
The fix was added to drake already, but the patch to director was held up until we updated drake master.

mauricefallon:
Thanks @patmarion , that works well and solves the primary problem.

I did notice that some of the scale tags actually use a negative scale. This is for links on the left and right half of a symmetical robot. I see that the drake parse doesn't support this so, I don't think you should do anything, but it can occur.

FWIW, drake_urdf_renderer.cpp (the old-style gtk renderer?) supports independent scales.

The example:
    <visual>
      <origin rpy="0 0 0" xyz="0 0 0"/>
      <geometry name="LThighUpLeg_visual">
        <mesh filename="package://bot_urdf/meshes/HipYaw.STL" scale="0.001 -0.001 0.001"/>
      </geometry>
      <material name="dark_grey"/>
    </visual>

patmarion:
I see.  The drake Mesh object only represented scale with a single value, so I made the parser work that way.  The one urdf file I looked at only had a scalar value for scale.  But now I see many urdf files have 3 components.  I made a start here, but I have not tested it yet.  (I'm not sure the ss.good() call does what I think it does)

https://github.com/RobotLocomotion/drake/compare/master...patmarion:support-three-component-scale

PR #134, opened 2015-11-30
Reduce dependencies

patmarion:


patmarion:
this is some work toward removing the dependencies on pcl/lcm/libbot/midi from the core python modules.  Still some remaining work in viewbehaviors.py.

PR #142, opened 2015-12-18
Revert "Specify LCM community for the video logger via environment viariable."

patmarion:
Reverts RobotLocomotion/director#141

### DIFF
--- a/src/python/scripts/videoLogServer.py
+++ b/src/python/scripts/videoLogServer.py
@@ -13,7 +13,8 @@ import numpy as np
 from director import lcmspy as spy
 
 
-VIDEO_LCM_URL = os.getenv('VIDEO_LCM_URL', 'udpm://239.255.76.50:7650?ttl=1')
+VIDEO_LCM_URL = 'udpm://239.255.76.50:7650?ttl=1'
+
 
 class FieldData(object):
 
@@ -248,7 +249,7 @@ class ServerThread(object):
 class LogSyncThread(object):
 
     def __init__(self, sharedUtimeMap):
-        print 'Creating control class'
+
         self.sharedUtimeMap = sharedUtimeMap
         self.utimes = None
         self.logLookup = LogLookup()
@@ -280,6 +281,7 @@ class LogSyncThread(object):
 
 
     def onFrameRequest(self, utimeRequest):
+
         if self.logLookup.utimeMap is None:
 
             self.logLookup.setUtimeMap(dict(self.sharedUtimeMap))
@@ -306,6 +308,7 @@ class LogSyncThread(object):
 
 
     def onControlMessage(self, channel, msgBytes):
+
         if self.getLastPublishElapsedTime() > self.publishFrequency:
             msg = spy.decodeMessage(msgBytes)
             self.onFrameRequest(msg.utime)


patmarion:
hang on, if we want to make this more flexible then there is a little more work required.  I think everything does work correctly without modifying files, i may have just been a little confused when i typed some instructions on the other github issue.

patmarion:
What was your preferred setting of VIDEO_LCM_URL?

mithrandir123:
I don't know yet. We'll have a URL for the robot machine when we set it up but I was using just my localhost/default url for this for now, i.e. 
```Shell
VIDEO_LCM_URL=""
```

I pulled it out into environment variable so that I can set it in the pmd file (Same as the other URL set in the video_capture.pmd).

patmarion:
VIDEO_LCM_URL is supposed to be private to the video capture and playback tools.  It was selected to be different than any other URL, because we don't want gigabytes of GoPro video LCM traffic to be in the robot lcm community.

The URL selected for VIDEO_LCM_URL is hard coded in that script and in the pmd file.  I'm happy to make it an environment variable instead, but then we should change it everywhere, and make it mandatory.  But I'm not sure it needs to be user settable, it's just an implementation detail of the video capture tools.

I think these are the issues:

- the pmd launches videoLogServer.py with LCM_DEFAULT_URL set to the same thing as VIDEO_LCM_URL.  That LCM_DEFAULT_URL that the videoLogServer.py tool is launched with is the url that it will use to listen for EST_ROBOT_STATE for playback syncing.  So as it is hard coded in the pmd, you must playback robot logs using VIDEO_LCM_URL for syncing to work, otherwise change the pmd, perhaps it should use the "review" url I mentioned yesterday.

- the videoLogPlayer.py is supposed to be launched with LCM_DEFAULT_URL set to VIDEO_LCM_URL, which is the url everything else is launched with in the pmd.  The videoLogPlayer communicates with videoLogServer over this URL..

Of course, documentation is lacking, sorry about that.  You are the first person to attempt to use these tools outside our lab, where we only have one computer equipped with a capture device, and everything was designed to just work on that computer and then no more time was spent to make these tools general purpose.

mithrandir123:
> the videoLogPlayer.py is supposed to be launched with LCM_DEFAULT_URL set to VIDEO_LCM_URL
This I didn't know but it makes sense now.
I got confused by the videoLogPlayer.py, because I started it without setting LCM_DEFAULT_URL. In which case I had to change the VIDEO_LCM_URL for  videoLogServer.py.

I'll run it again for my own sanity.
Where do you usually launch the videoLogPlayer.py? With the director or with the video logger?

patmarion:
On our setup, videoLogPlayer.py is launched using a gnome task bar launcher.  It is launched using directorPython and you should set LCM_DEFALUT_URL to VIDEO_LCM_URL ( udpm://239.255.76.50:7650?ttl=1 )

patmarion:
I just confirmed the tools are working without modifications, so I'm going to merge this revert.

Everything works as long as you start videoPlayer.py on the correct LCM url.

I think it would be beneficial to remove the hardcoded lcm url as you suggested and replace it with an environment variable.  In this case, I think the environment variable should be mandatory, and used in both videoLogServer.py and videoPlayer.py.

PR #148, opened 2016-01-12
Update final pose planning

MarcoCar:
Update the interface to the new version of the final pose planner in matlab using point clouds to detect collisions

### DIFF
--- a/src/python/director/ikplanner.py
+++ b/src/python/director/ikplanner.py
@@ -87,7 +87,17 @@ class ConstraintSet(object):
             nominalPoseName = self.startPoseName
 
         ikParameters = self.ikPlanner.mergeWithDefaultIkParameters(self.ikParameters)
-        self.plan = self.ikPlanner.runIkTraj(self.constraints, self.startPoseName, self.endPoseName, nominalPoseName, ikParameters=ikParameters)
+
+        if getIkOptions().getPropertyEnumValue('Use collision') == 'RRT*':
+            mapServerSource = om.findObjectByName('Map Server').source
+            pointCloud = mapServerSource.getOctreeWorkspaceData()
+            if pointCloud == (None, None):
+                self.plan = []
+                print 'No point cloud data found'
+            else:
+                self.plan = self.ikPlanner.runIkTraj(self.constraints, self.startPoseName, self.endPoseName, nominalPoseName, ikParameters=ikParameters, pointCloud = pointCloud)
+        else:
+            self.plan = self.ikPlanner.runIkTraj(self.constraints, self.startPoseName, self.endPoseName, nominalPoseName, ikParameters=ikParameters)
 
         return self.plan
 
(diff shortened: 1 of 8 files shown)


patmarion:
Is this still work in progress, or ready for review?

mauricefallon:
Marcos is still working on it

MarcoCar:
@patmarion @mauricefallon this is now ready for review

src/python/director/perception.py
```diff
@@ -518,6 +518,22 @@ def showMap(self, viewId, mapId):
     def getSceneHeightData(self):
         return self.getDepthMapData(lcmdrc.data_request_t.HEIGHT_MAP_SCENE)
 
+    def getOctreeWorkspaceData(self):
+        return self.getPointCloudData(lcmdrc.data_request_t.OCTREE_WORKSPACE)
+
+    def getPointCloudData(self, viewId):
```
patmarion: I'd prefer this method return the polyData object.  The caller can extract the points if they want.

src/python/director/teleoppanel.py
```diff
@@ -341,12 +345,13 @@ def updateIk(self):
 
 
     def updateCollisionEnvironment(self):
-        affs = self.panel.affordanceManager.getCollisionAffordances()
-        if not affs:
-            self.panel.ikPlanner.ikServer.clearEnvironment()
-        else:
-            urdfStr = affordanceurdf.urdfStringFromAffordances(affs)
-            self.panel.ikPlanner.ikServer.setEnvironment(urdfStr)
+        if ~self.getCheckboxState(self.ui.finalPosePlanningOptions):
```
patmarion: this is using a matlab style ~ for not

src/python/director/teleoppanel.py
```diff
@@ -447,14 +452,18 @@ def updateConstraints(self):
             elif self.getLFootConstraint() == 'constrained':
                 constraints.extend(ikPlanner.createSixDofLinkConstraints(startPoseName, ikPlanner.leftFootLink, tspan=[1.0, 1.0]))
             elif self.getLFootConstraint() == 'sliding':
-                constraints.extend(ikPlanner.createSlidingFootConstraints(startPoseName)[:2])
+                if not self.getCheckboxState(self.ui.finalPosePlanningOptions):
+                    constraints.extend(ikPlanner.createSlidingFootConstraints(startPoseName)[:2])
```
patmarion: this logic is kind of strange.  Maybe the foot combos should have a new value called 'feet together' instead of overloading the sliding option?

src/python/director/ikplanner.py
```diff
@@ -22,6 +23,7 @@
 from director import planplayback
 from director import segmentation
 from director import drcargs
+from ddapp.perception import MapServerSource
```
patmarion: looks like this import isn't used, can we remove it?

src/python/director/ikplanner.py
```diff
@@ -7,6 +7,7 @@
 import functools
 import random
 import numpy as np
+from vtk.util.numpy_support import vtk_to_numpy
```
patmarion: looks like this import isn't used, can we remove it?

patmarion:
Looks good overall!  My general comment is that I'm not thrilled was adding special case pointcloud logic to runIkTraj, and building more custom UI in teleop panel.  It would take more design work, but it would be fantastic to have a separate Collision Free Planning panel that self contained all this work and the ik.py file could remain general purpose.  I'm happy to go ahead with this pull request, but if there is more development in this direction I think it would be worth investing the time to separate the work into a new UI and planner (so you don't have to modify teleoppanel.py and ik.py)

MarcoCar:
Thanks @patmarion. I'll follow your suggestion and make a new pull request when the new panel is ready.

PR #152, opened 2016-01-18
import iris earlier to avoid import error

patmarion:
On some systems we are getting the import error:

ImportError: dlopen: cannot load any more object with static TLS

which seems to be coming from the swig layer of iris.  The issue is
avoided by importing iris earlier.  It seems to be interacting in a
strange way with other modules loaded by director.


rdeits:
Okay, that's interesting. It appears that it's possibly a result of some library being statically linked somewhere within the IRIS dependencies. My guess would be the `cdd` library, since that wasn't previously available from Python before the swig conversion. If this continues to be an issue, I can investigate further.

PR #153, opened 2016-01-18
Make Valkyrie the default to load

iamwolf:


patmarion:
I'm happy to make this switch.  I think it might break a few commands in our atlas pmd (for driving) that assume atlas is the default robot, but we're not using that anymore.  I'll fix those as they are discovered.  Also, several of the unit tests load atlas.  This may break some tests.  I'll merge this and let the build server report errors and we can fix them if they appear.

PR #154, opened 2016-01-18
Flip imageOverlay camera based on head frame

iamwolf:
Reads the roll of the head frame and rotates the imageOverlay accordingly (to automatically account for the Valkyrie/Atlas Multisense rotation for instance). Next up is the same for the Camera View field.

### DIFF
--- a/src/python/director/cameraview.py
+++ b/src/python/director/cameraview.py
@@ -417,6 +417,7 @@ class CameraImageView(object):
 
         imageManager.addImage(imageName)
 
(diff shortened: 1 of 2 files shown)


iamwolf:
I'm actually changing this to a rotation in ImageManager at the moment and will update with a comment once it's ready to review

mauricefallon:
sorry, i missed the newer comment. can you fix this this up and merge when you are ready?

iamwolf:
Yes, will do

PR #157, opened 2016-01-18
Flip images in ImageManager if camera frame is rotated

iamwolf:
This flips the images directly in the ImageManager if the camera frame is rotated. Rotating at the lowest level is preferred compared to just rotating the camera when displaying it (and cascades through every use of the image). Tested with val2 and v5.

### DIFF
--- a/src/python/director/cameraview.py
+++ b/src/python/director/cameraview.py
@@ -211,6 +211,7 @@ class CameraView(object):
 
         self.imageManager = imageManager
         self.updateUtimes = {}
+        self.robotModel = None
         self.sphereObjects = {}
         self.sphereImages = [
                 'CAMERA_LEFT',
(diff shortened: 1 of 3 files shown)


src/python/director/startup.py
```diff
@@ -1077,6 +1069,14 @@ def stop(self):
     if (drcargs.args().exo):
         ikPlanner.pushToMatlab = False
 
+# Rotate Multisense image/CAMERA_LEFT if the camera frame is rotated (e.g. for Valkyrie)
+if robotStateModel.getHeadLink():
+    tf = robotStateModel.getLinkFrame(robotStateModel.getHeadLink())
+    roll = transformUtils.rollPitchYawFromTransform(tf)[0]
+    if np.isclose(np.abs(roll), np.pi, atol=1e-1):
+        cameraview.imageManager.setImageRotation180('CAMERA_LEFT')
+
```
patmarion: I think this could by tidier.  This code can be moved up to where cameraview.init() is called, and look like:

cameraview.initImageRotations(robotStateModel)

src/python/director/filterUtils.py
```diff
@@ -114,3 +114,18 @@ def removeNonFinitePoints(polyData, arrayName='Points'):
...
+    '''
+    r1 = vtk.vtkImageFlip()
+    r1.SetInput(image)
+    r1.SetFilteredAxis(0)
+    r1.Update()
+    r2 = vtk.vtkImageFlip()
+    r2.SetInput(r1.GetOutput())
+    r2.SetFilteredAxis(1)
+    r2.Update()
+    return r2.GetOutput()
```
patmarion: this should say:  return shallowCopy(r2.GetOutput())

(you will still need the ShallowCopy call in your calling code)

src/python/director/cameraview.py
```diff
@@ -132,6 +133,7 @@ def addImage(self, name):
         self.imageUtimes[name] = 0
         self.images[name] = image
         self.textures[name] = tex
+        self.imageRotations180[name] = None
```
patmarion: This could be a set instead of a dict.  If you prefer a dict of booleans, maybe initialize to False instead of None?

patmarion:
Looks great! See inline comments.

iamwolf:
@patmarion thanks for the feedback - I've modified the PR accordingly

PR #158, opened 2016-01-19
Deactivate Driving Planner (was throwing errors)

iamwolf:
Driving Planner is throwing warnings/errors on each director launch. Given that we are probably not going to drive with the robots any time soon, does it make sense to disable it from loading by default?


patmarion:
what are the warnings/errors? Not sure if I have seen it on my computer.

iamwolf:
Oops, this was local to my machine (possibly after an expired Gurobi license - now resolved). Apologies for wrongly flagging it up

PR #160, opened 2016-01-20
Improved texture support

patmarion:


RussTedrake:
Thanks @patmarion.
@Sproulx uses the precompiled directory binaries on windows.  Perhaps he can just manually copy in these py files?

patmarion:
I'll post modified files.  Copying these directly will bring in some other changes that won't be compatible with the windows version since those are rather old.

Maybe we can open an issue if one doesn't already exist to get a Windows build server with VTK 5.10, Qt 4.8, Python 2.7 installed, then we can build director on Windows.  I have these dependencies installed on my Windows VM, but it's not straightforward to install these dependencies on a new Windows computer except from source.

RussTedrake:
It’s on the roadmap (implicitly).  Feel free to open an issue, but we won’t forget.  :)

patmarion:
@Sproulx  I have put a new drake-v0.9.11-win64 here:

    https://www.dropbox.com/s/k4zx378yh3e946d/drake-v0.9.11-win64.zip?dl=0

But all you really need are these two files from inside the zip:

build/lib/python2.7/dist-packages/ddapp/ioUtils.py
build/lib/python2.7/dist-packages/ddapp/drakevisualizer.py

You can copy those two files from the folder I have provided and place them in the same location in your drake folder.

RussTedrake:
Thanks Pat.  Can i recommend you post this as a release on the director pages?  
I’d like to add a precompiled director as a make option for the external project for drake, too.

Sproulx:
Thanks, @patmarion.

Sproulx:
I can now render Val with textures in the visualizer. Unfortunately, when I try to render the prius model I've been working on, it still comes up with out textures, regardless if I list a dae or obj in the urdf. I know that the Val urdf uses dae files for meshes with texture. Is it stil converting those to obj? I get an error, in Matlab, for failed conversion to obj if I don't have an obj of the same name as the dae in the containing folder.

patmarion:
That's great.  Let's discuss the rest over at https://github.com/RobotLocomotion/drake/issues/1583

PR #165, opened 2016-01-26
Sn continuous walking.

simalpha:
Fixed continuous walking. Added save to .vtk option.

### DIFF
--- a/src/python/director/continuouswalkingdemo.py
+++ b/src/python/director/continuouswalkingdemo.py
@@ -5,6 +5,7 @@ import operator
 import vtkAll as vtk
 import vtkNumpy
 import functools
+import ihmc
 
 from director import lcmUtils
 from director import ioUtils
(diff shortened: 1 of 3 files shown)


src/python/director/continuouswalkingdemo.py
```diff
@@ -151,7 +152,7 @@ def _setupOnce(self):
         #footContactSubContinuous.setSpeedLimit(60)
 
         lcmUtils.addSubscriber('FOOTSTEP_PLAN_RESPONSE', lcmdrc.footstep_plan_t, self.onFootstepPlanContinuous)# additional git decode stuff removed
-        lcmUtils.addSubscriber('IHMC_FOOTSTEP_STATUS', lcmdrc.footstep_status_t, self.onFootstepStatus)
+        lcmUtils.addSubscriber('IHMC_FOOTSTEP_STATUS', ihmc.footstep_status_t, self.onFootstepStatus)
```
patmarion: did this message actually move out of drc lcmtypes?

simalpha: Currently, there is no footstep_status_t message in drc. This message was renamed from ipab (https://github.com/openhumanoids/oh-distro/commit/82e2fac78efd31d89d312087f821430b0d607691). Instead of using this message, I can look for another solution.

patmarion: got it, thanks!

PR #166, opened 2016-01-26
use thirdparty transformations.py instead of botpy

patmarion:


patmarion:
Opening this for review.  @mauricefallon this modified transformUtils.py to remove the dependency on libbot's rotations library.  Last year I added thirdparty.transformations.py and a test that verified it returns the same answers as libbot rotations, but I did not actually switch to using it inside the transformUtils.py module which is used everywhere in director.  I think this pull request is an improvement because it removes a dependency without removing any functionality, but I'm curious what you think.

patmarion:
@mauricefallon can you let me know if you have any concerns with this going in?  I don't expect that this should have any effect on transform results or performance, and it has an associated test, but still we should watch out for rotation or transform bugs in case there is something wrong.

mauricefallon:
Would you mind me running some tests myself? I remember Dehann suggesting libbot and drake gave different transformation outputs in corner case ... so im sensitive to these kinds of things.

patmarion:
Yes, please go ahead.  The transformations test is already in master, I added it last year.  You could add new tests to this file.  You can run it with this command:

    cddrc
    directorPython director/src/python/tests/testTransformations.py

This file, testTransformations.py, tests that the answers from thirdparty.transformations agree with libbot and director's transformUtils.py.  Director's transformUtils.py currently uses libbot internally, but after this pull request it will use thirdparty.transformations internally.

patmarion:
I am waiting on this pull request so that I can make an update to drake-visualizer.  Let me know if you might have a chance to review this today?

mauricefallon:
yes, I'll look at it next. about an hour or two.

src/python/director/transformUtils.py
```diff
@@ -17,13 +17,9 @@ def getTransformFromNumpy(mat):
...
     '''
-    m = vtk.vtkMatrix4x4()
-    for r in xrange(4):
-        for c in xrange(4):
-            m.SetElement(r, c, mat[r][c])
-
+    assert mat.shape == (4,4)
     t = vtk.vtkTransform()
-    t.SetMatrix(m)
+    t.SetMatrix(mat.flatten())
```
patmarion: This is a speedup, no change in behavior.

src/python/director/transformUtils.py
```diff
@@ -149,7 +145,7 @@ def frameInterpolate(trans_a, trans_b, weight_b):
     [pos_a, quat_a] = poseFromTransform(trans_a)
     [pos_b, quat_b] = poseFromTransform(trans_b)
     pos_c = pos_a *(1-weight_b) + pos_b * weight_b;
-    quat_c = botpy.quat_interpolate(quat_a,quat_b, weight_b)
+    quat_c = transformations.quaternion_slerp(quat_a, quat_b, weight_b)
```
patmarion: This replaces botpy.quat_interpolate with transformations.quaternion_slerp.  The associated test confirms that both functions give the same result (using randomly selected quaternions).

src/python/director/transformUtils.py
```diff
@@ -157,28 +153,18 @@ def transformFromPose(position, quaternion):
     '''
     Returns a vtkTransform
     '''
-    rotationMatrix = np.zeros((3,3))
-    vtk.vtkMath.QuaternionToMatrix3x3(quaternion, rotationMatrix)
 
-    mat = np.eye(4)
-    mat[:3,:3] = rotationMatrix
+    mat = transformations.quaternion_matrix(quaternion)
```
patmarion: This replaces VTK's QuaternionToMatrix3x3 with the equivalent conversion routine provided by transformations.py

src/python/director/transformUtils.py
```diff
@@ -157,28 +153,18 @@ def transformFromPose(position, quaternion):
...
     mat[:3,3] = position
-
-    t = vtk.vtkTransform()
-    t.SetMatrix(mat.flatten())
-    return t
+    return getTransformFromNumpy(mat)
```
patmarion: code re-use

src/python/director/transformUtils.py
```diff
@@ -157,28 +153,18 @@ def transformFromPose(position, quaternion):
...
     Returns position, quaternion
     '''
-    angleAxis = range(4)
-    transform.GetOrientationWXYZ(angleAxis)
-    angleAxis[0] = math.radians(angleAxis[0])
-    pos = transform.GetPosition()
-    quat = botpy.angle_axis_to_quat(angleAxis[0], angleAxis[1:])
-    return np.array(pos), np.array(quat)
+    mat = getNumpyFromTransform(transform)
+    return np.array(mat[:3,3]), transformations.quaternion_from_matrix(mat, isprecise=True)
```
patmarion: Previous method:  VTK transform to angle axis representation followed by botpy.angle_axis_to_quat
New method: call transformations conversion routine for matrix to quaternion.

The associated test confirms this provides the same answer for a randomly selected transform.

src/python/director/transformUtils.py
```diff
@@ -186,23 +172,18 @@ def frameFromPositionAndRPY(position, rpy):
...
-
-    t = vtk.vtkTransform()
-    t.PostMultiply()
-    t.RotateWXYZ(math.degrees(angle), axis)
-    t.Translate(position)
-    return t
+    rpy = np.radians(rpy)
+    mat = transformations.euler_matrix(rpy[0], rpy[1], rpy[2])
+    mat[:3,3] = position
+    return getTransformFromNumpy(mat)
```
patmarion: Previous method: uses botpy's RPY to angle axis, then VTK's angle axis to homogeneous transform
New method: use rpy to homogeneous transform conversion from transformations.py

src/python/director/transformUtils.py
```diff
@@ -186,23 +172,18 @@ def frameFromPositionAndRPY(position, rpy):
...
 
 
 def rollPitchYawToQuaternion(rpy):
-    return botpy.roll_pitch_yaw_to_quat(rpy)
+    return transformations.quaternion_from_euler(rpy[0], rpy[1], rpy[2])
```
patmarion: replaces botpy with transformations.py

src/python/director/transformUtils.py
```diff
@@ -186,23 +172,18 @@ def frameFromPositionAndRPY(position, rpy):
...
 
 
 def quaternionToRollPitchYaw(quat):
-    return botpy.quat_to_roll_pitch_yaw(quat)
+    return transformations.euler_from_quaternion(quat)
```
patmarion: replaces botpy with transformations.py

mauricefallon:
Apologies for the big delay, but I'm happy to accept this.

You probably know this, these methods are actually used in ROS-python:
https://github.com/ros/geometry/tree/indigo-devel/tf/src/tf
http://wiki.ros.org/geometry/RotationMethods
Where their version is over 7 years old.

The calculation in libbot and transformation.py are not the same but happy to reply on this well used tool

patmarion:
Ok, sounds good.  Merging.

The version in ROS tf is from 2009.  The transformations module is actively maintained outside of ROS.  The current version is 2015.07.18, and that is the version used by director.

mauricefallon:
Agreed, the changes since 2009 are very minimal - I diff'ed them.
Which is to suggest that its "right"

patmarion:
You said you might have an email which describes some rotations in libbot that would be worth checking against other transform libraries?  If you forward that then I can look into it at some point.

mauricefallon:
found it:
https://github.com/openhumanoids/oh-distro-private/issues/419

seems like a pretty clean issue to check

patmarion:
I just tested and transformations.py gives the correct answer listed in that issue, which is not the answer provided by libbot (that issue claims libbot produces the wrong answer).  I think the two libraries agree most places, but that issue highlights an edge case where libbot's behavior is not correct.

mauricefallon:
It was Dehann's assertion that it's 'correct' but I don't agree.

Eigen produces a 3rd different answer:
0
0.707107
0
-0.707107

... and then converting Eigen back to a rotation matrix produces some eps values to the original:
 2.22045e-16            0           -1
           0           -1           -0
          -1            0 -2.22045e-16

This matrix is of course a singularity. For example libbot has a special check as the trace of the 4x4 matrix is zero.

@tkoolen : could you suggest some intuition: how can this particular rotation matrix produce some many different quaternions?

Eigen Code snippet:

    Eigen::Matrix3d y;
    y << 0.,0.,-1.,   0.,-1.,0.,   -1.,0.,0.;
    Eigen::Quaterniond quatE(y);
    std::cout << "\n";
    std::cout << quatE.w() << "\n";
    std::cout << quatE.x() << "\n";
    std::cout << quatE.y() << "\n";
    std::cout << quatE.z() << "\n";

    Eigen::Matrix3d y2 = Eigen::Matrix3d(quatE);
    std::cout << y2 < "\n";

patmarion:
The Eigen answer is actually the same as Dehann's assertion.  If you negate each component of the quaternion you have an equivalent representation of the same rotation.

tkoolen:
Yeah, @patmarion is right about that.

I also think that Dehann's assertion was correct. 

While there are various ways to go from rotation matrix to quaternion, there's really only one way to go from quaternion to rotation matrix, and it doesn't involve any if statements or cases.

So since the conversion from quaternion to rotation matrix is fixed, converting the quaternion you got from rotation matrix --> quaternion back into a rotation matrix and checking that the resulting rotation matrix is the same as what you started with is a good test of the rotation matrix --> quaternion conversion.

Dehann reports that libbot returns this quaternion:
```
quat_libbot = [0;
0.7071;
0;
0.7071];
```

Drake's Matlab quat2rotmat function tells us that
```
quat2rotmat(quat_libbot) = 
[0 0 1;
0 -1 0;
1 0 0]
```
which is just plain wrong.

I also checked against https://en.wikipedia.org/wiki/Rotation_matrix#Conversion_from_and_to_axis-angle, which uses an intuitive eigenvalue method, and found the same result as Eigen (up to negation).

mauricefallon:
That's great to know. I'm happy to agree that its libbot that's incorrect. Thanks!

(In the future it would be a goal to re-implement libbot using Eigen as the basis of it's transformations)

BTW, transformations.py does also have some code to take core of numerical precision (see 'isprecise' in quaternion_from_matrix).

PR #168, opened 2016-01-28
extract atlas types

mauricefallon:
https://github.com/openhumanoids/oh-distro/pull/32

### DIFF
--- a/src/python/director/continuouswalkingdemo.py
+++ b/src/python/director/continuouswalkingdemo.py
@@ -24,6 +24,7 @@ import director.terrain
 import director.tasks.robottasks as rt
 
 import drc as lcmdrc
+import atlas
 
 from thirdparty import qhull_2d
 from thirdparty import min_bounding_rect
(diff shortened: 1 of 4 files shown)


src/python/director/continuouswalkingdemo.py
```diff
@@ -153,7 +153,7 @@ def _setupOnce(self):
         lcmUtils.addSubscriber('FOOTSTEP_PLAN_RESPONSE', lcmdrc.footstep_plan_t, self.onFootstepPlanContinuous)# additional git decode stuff removed
         lcmUtils.addSubscriber('IHMC_FOOTSTEP_STATUS', lcmdrc.footstep_status_t, self.onFootstepStatus)
         lcmUtils.addSubscriber('EST_ROBOT_STATE', lcmdrc.robot_state_t, self.onRobotStatus)
-        stepParamsSub = lcmUtils.addSubscriber('ATLAS_STEP_PARAMS', lcmdrc.atlas_behavior_step_params_t, self.onAtlasStepParams)
+        stepParamsSub = lcmUtils.addSubscriber('ATLAS_STEP_PARAMS', atlas.behavior_step_params_t, self.onAtlasStepParams)
```
patmarion: Does this file need an import atlas?  Same for some of the other files?

mauricefallon: I guess it imports atlasdriver.py which imports atlas
I did a clean build and fixed up the errors that came up, so i think this is sufficient

mauricefallon: ok to accept this?

patmarion: I haven't had a chance to try running it, but I'm pretty sure you need to "import atlas" in all the files, not just atlasdriver.py.  It won't throw errors until the code is executed, and not all of this code is executed on start up so you wouldn't notice it right away.

mauricefallon: I've just added the imports. I think its ok now

PR #170, opened 2016-02-01
Add camera control panel

patmarion:


mauricefallon:
looks great, can you give a video of what controls this gives?
Seems to add a 'chase cam'?

patmarion:
Yep, chase camera is one of the available modes.  I'll work on making a video to demonstrate.

Currently objects with frames are supported, but robot urdf objects are not.  I will need to write an adapter for robot models- it would just simply extract the floating base joint as a transform.  Supporting robots seems like an obvious feature, but I initially developed this for drake-visualizer which doesn't have robot urdfs.

The camera control panel has some UI for selecting a target object and then switching between camera control modes.  There is a general interface to write your own camera control mode, and several modes are provided:

  - track position
  - track position & orientation
  - smooth follow (chase cam)
  - look at (camera on a tripod always points at target)
  - orbit (camera orbits target while keeping target centered)

PR #173, opened 2016-02-02
Fix unit tests for Valkyrie

iamwolf:
This addresses and resolves https://github.com/openhumanoids/oh-distro/issues/22

### DIFF
--- a/src/python/tests/testTeleopPanel.py
+++ b/src/python/tests/testTeleopPanel.py
@@ -34,7 +34,7 @@ def checkGraspFrame(inputGraspFrame, side):
 def onIkStartup(ikServer, startSuccess):
 
     side = 'left'
-    goalFrame = transformUtils.frameFromPositionAndRPY([0.5, 0.7, 1.2], [0, 90, -90])
+    goalFrame = transformUtils.frameFromPositionAndRPY([0.5, 0.5, 1.2], [0, 90, -90])
 
     assert not checkGraspFrame(goalFrame, side)
     frame = teleopPanel.endEffectorTeleop.newReachTeleop(goalFrame, side)
--- a/src/python/tests/testContinuousWalking.py
+++ b/src/python/tests/testContinuousWalking.py
@@ -30,7 +30,7 @@ def processSingleBlock(robotStateModel, whichFile=0):
     else:
         polyData = ioUtils.readPolyData(os.path.join(dataDir, 'terrain/block_top.vtp'))
 
-    standingFootName = 'l_foot'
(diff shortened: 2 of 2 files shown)


iamwolf:
Note: Both tests can now be run in a matrix using both default (valkyrie) and -v5 (working for both)

src/python/tests/testTeleopPanel.py
```diff
@@ -34,7 +34,7 @@ def checkGraspFrame(inputGraspFrame, side):
 def onIkStartup(ikServer, startSuccess):
 
     side = 'left'
-    goalFrame = transformUtils.frameFromPositionAndRPY([0.5, 0.7, 1.2], [0, 90, -90])
+    goalFrame = transformUtils.frameFromPositionAndRPY([0.5, 0.5, 1.2], [0, 90, -90])
```
patmarion: I guess this is a goal frame that is reachable by both robots? nice!

patmarion:
I think I updated src/python/tests/CMakeLists.txt to use -v5 for these tests.  Keeping that for now?

iamwolf:
Didn't see that (https://github.com/RobotLocomotion/director/commit/7397749660b9a2d93cb0dbba53018258ab4da571). Happy to go with what you consider best. The changes here won't have an effect to current behaviour, rather allow being run for both v5 and val2. Is there a way to edit the CMakeLists to run these tests with the extra argument and without?

src/python/tests/testTeleopPanel.py
iamwolf: Indeed. It returns 13 for Valkyrie

patmarion:
I think the CMakeLists.txt uses a for loop that makes it tricky to repeat a test twice, but we could extend it to support this use case, would be nice to do.

iamwolf:
I'm happy to have a go at it. What would be the best approach for this? Is expanding the add_python_test macro with a check whether extra_args is set, and if not adding it additionally with the extra argument v5, an ok solution or is there a preferable way in CMake?

patmarion:
Right now the for loop is designed to add one test for each listed test file.  It {test name}_extra_args is defined then those args are added to the test command line.

Maybe we could design add_python_test() like this:

  - If {test name}_extra_args is defined, add it to the test command line
  - if {test name}_extra_args_2 is defined, add a 2nd test named {test name}_2 with the 2nd args
  - keep looping, check for {test name}_extra_args_{n} until the variable is no longer defined

So then we could have:

set(testTeleopPanel_extra_args -v5)
set(testTeleopPanel_extra_args_2 -val2)

And this would add two tests, one with -v5 and one with -val2.

PR #175, opened 2016-02-03
Travis ci build

patmarion:


iamwolf:
:+1:

.travis.yml
```diff
@@ -0,0 +1,49 @@
...
+
+notifications:
+  email: false
+
+branches:
+  only:
+    - master
+
+before_install:
+  - bash ./distro/travis/before_install.sh
```
rdeits: @patmarion I remembered why I broke off the drake pre-build contents into their own script: Travis can't handle multi-line bash IF statements inside `.travis.yml`: https://github.com/travis-ci/travis-ci/issues/497 https://groups.google.com/forum/#!topic/travis-ci/uaAP9zEdiCg 

But putting them inside a `.sh` file and then calling that script works just fine.

distro/travis/before_install.sh
```diff
@@ -0,0 +1,7 @@
+if [ "$TRAVIS_OS_NAME" = "osx" ]; then
```
rdeits: You might want to `set -e` before this so that the script will fail if any line fails

patmarion:
Thanks for the review!  This PR is not ready to be merged.  This is just the very first introduction of the .travis.yml to kick off the first travis build.  I plan to push commits to this PR while I debug and get all the features working.

PR #176, opened 2016-02-03
Travis ci build

patmarion:


CTestConfig.cmake
```diff
@@ -0,0 +1,7 @@
+set(CTEST_PROJECT_NAME "DRC")
+set(CTEST_NIGHTLY_START_TIME "22:00:00 EDT")
+
+set(CTEST_DROP_METHOD "http")
+set(CTEST_DROP_SITE "kobol.csail.mit.edu")
+set(CTEST_DROP_LOCATION "/cd/submit.php?project=DRC")
```
iamwolf: Should this be changed to project=Director before merging?

patmarion: yes, thanks!  Still need to add it.

PR #177, opened 2016-02-03
add color octrees

mauricefallon:
this is towards an immerse 3D environment in the UI - with history.
(In parallel @simalpha  is making great progress with using libpointmatcher from ETHZ)

All the added cpp and h files are directly from the up-to-date version of octomap's own visualiser in externals (octovis, with one function removed).

This doesn't (yet) expose controls to the user but it can:
- accept octrees over lcm in both binary and color formats
- read both kinds from file and then draw them

This shows switching between binary and color octrees:
https://www.dropbox.com/s/fz6fu6yptkmhq7g/from_lcm_mono_then_color.mp4?dl=0
(the rainbow colors are stored in the octree LCM message)

This (poor quality) large model can be rendered, but needs to be read from file:
https://www.dropbox.com/s/a56hbjl5dxb1k19/from_file.mp4?dl=0

CMU had this visualisation mode for Chimp:
https://youtu.be/In8unM_zJr4?t=7


mauricefallon:
(comment update online)

patmarion:
Cool, nice update.

Is it necessary to add the files from octovis?  It looks like the octomap external project installs liboctovis.so and headers, could you use those?

PR #178, opened 2016-02-04
Work without headLink

iamwolf:
Currently when the directorConfig does not have a headLink, director fails to fully load since cameraview is trying to load one. This will return None if there is no headLink in the director config.

### DIFF
--- a/src/python/director/roboturdf.py
+++ b/src/python/director/roboturdf.py
@@ -106,7 +106,7 @@ class RobotModelItem(om.ObjectModelItem):
             return None
 
     def getHeadLink(self):
-        return drcargs.getDirectorConfig()['headLink'] if 'headLink' in drcargs.getDirectorConfig() else None
+        return drcargs.getDirectorConfig().get('headLink')
 
     def getLinkContactPoints(self, linkName):
         pts = self.model.getBodyContactPoints(linkName)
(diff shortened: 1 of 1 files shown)


src/python/director/roboturdf.py
```diff
@@ -106,7 +106,7 @@ def getLinkFrame(self, linkName):
             return None
 
     def getHeadLink(self):
-        return drcargs.getDirectorConfig()['headLink']
+        return drcargs.getDirectorConfig()['headLink'] if 'headLink' in drcargs.getDirectorConfig() else None
```
patmarion: or maybe:

    return drcargs.getDirectorConfig().get('headLink')

patmarion:
Looks good.  Maybe you want to address the in-line comment and then I'll merge.

However, robotModel.getHeadLink() is a bit of a hack.  We have a few things like this hanging around that would nice to clean up eventually.  I'd be curious to hear ideas if someone wants to propose a more general interface for querying links in a general way.  I think the directorConfig works ok but I'm sure we could improve on it.

iamwolf:
Thanks, comment addressed. It's true that it is a hack and it returning 'None' might as well lead other bits and pieces of the code to break if they don't check the return but use them blindly. Happy to fix these as we go and discover them.

PR #179, opened 2016-02-04
add color octree minimal

mauricefallon:
Replacing:
https://github.com/RobotLocomotion/director/pull/177

This its possible to use octovis without modification.

But I have a few CMake issues. Could you suggest fixes?
- octovis doesn't install a pkgconfig file
- octovis requires something called QGLViewer - which is used only in one place - to draw an axes.
It is actually part of octomap and provides as FindQGLViewer.cmake
oh-distro/software/externals/octomap/octovis/src/extern/QGLViewer

I'll also PR some associated changes to open-humanoids if you want them

### DIFF
--- a/src/python/director/lcmoctomap.py
+++ b/src/python/director/lcmoctomap.py
@@ -22,9 +22,12 @@ class OctomapObject(om.ObjectModelItem):
         self.actor = actor
         self.actor.SetUseBounds(False)
         self.addProperty('Visible', actor.GetVisibility())
-        self.addProperty('Color Mode', 0, attributes=om.PropertyAttributes(enumNames=['Drab', 'Height']))
-        self.addProperty('Alpha', 0.8,
-                         attributes=om.PropertyAttributes(decimals=2, minimum=0, maximum=1.0, singleStep=0.1, hidden=False))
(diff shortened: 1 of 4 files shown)


mauricefallon:
The future direction:
https://www.dropbox.com/s/0gadcjeqadmqb0k/finals-map.mp4?dl=0
(the colors are mostly horrible at present)

patmarion:
very awesome!

I'm happy to help with the cmake code, I might not get to it right away, though, ok to leave this open a little while longer?

mauricefallon:
take as long as you need!

patmarion:
I just pushed a commit onto this branch to cleanup the cmakelists.

I have a commit to octovis that makes it possible to include octovis headers even though qglviewer is not installed.  I opened a pull request upstream here:

https://github.com/OctoMap/octomap/pull/108

In the meantime, we can change the url of our octomap external to point to my commit.  What do you think?

mauricefallon:
Your SHA is probably on top octomap/master ? 
But this change is unlikely to need such testing.

Go ahead and accept both this and the oh-distro PR, we can revert if necessary

Thanks for the CMake clean up!

PR #186, opened 2016-02-10
add cmake debugging messages

patmarion:
just for testing, not to be merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 os:
   - linux
-  - osx
+#  - osx
 
 language: generic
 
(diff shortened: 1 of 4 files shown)


iamwolf:
Pat, is the issue that travis can't find numpy? We had this issue for oh-distro as well and it took quite some time to debug it. The solution is as follows:
- add language: generic
- add: virtualenv: system_site_packages: true

patmarion:
Yes, that's the problem.  Thanks!  I will try that now.

I am very confused about it though, at cmake configure time, Director cannot import numpy.  But at test time, all the director tests pass, and those tests are able to import numpy.  Both of these events happen in the travis script: block.

patmarion:
I guess the tests run with directorPython instead of python, and that somehow inherits the system python path rather than the virutalenv path.

PR #195, opened 2016-02-15
point cloud receiver can accept rgb and float arrays

mauricefallon:
rgb color applied to color point cloud, float arrays seems to need to be rescaled


mauricefallon:
@patmarion : this allows attributes of a pointcloud to be received and added to a vtkPolyData. Any suggestions?

related to https://github.com/openhumanoids/oh-distro/pull/71

PR #196, opened 2016-02-16
Add package path

patmarion:


mauricefallon:
Can you explain what you are doing here? Does this change the interpretation of how urdfs are read or does this only impact on drake visualizer?

For example we had a list of directories to search for parts of Atlas, is that they effected?

patmarion:
Sure.  This pull request moves some code and adds functionality to drake-visualizer.  There are no changes to behavior for Director.  Here's some more info-

When Director starts it calls roboturdf.setupPackagePaths().  This function does two things:

1. it adds a list of pre-determined package paths in the software/models directory
2. it recursively searches directories listed in the ROS_PACKAGE_PATH environment variable for more packages

The directories added by step 1 take precedence.  Usually ROS_PACKAGE_PATH is empty, but it's useful it you are opening urdfs outside the software/models directory.

This pull request moved the code for step 2 into a standalone file named packagepaths.py so that it can be used reused more easily by drake-visualizer and other applications.  It also adds a convenience function to do filename resolution.  So you can do something like:

    m = PackageMap()
    m.populateFromEnvironment(['ROS_PACKAGE_PATH'])
    filename = m.resolveFilename('package://package_name/file.obj')

PR #204, opened 2016-02-27
move to using bot_core types for many messages

mauricefallon:
because of the python, I cannot be sure i captured all type imports.
I tested with both SCS and MIT simulators - walking plans and reaching plans.


patmarion:
looks good, go ahead and merge this when you're ready to merge the oh-distro branch too

PR #209, opened 2016-03-04
Fix testAtlasCommand.py after running on Val

iamwolf:


patmarion:
Great!  Are you finding this program useful?

By the way, if you want, we could rename this script, and maybe move it to python/scripts instead of python/tests.

iamwolf:
Yes, really useful to debug the LCM2ROSControl and to have a quick joint control interface. Will possibly add other features to it once needed :). Happy to keep it where it is for now or to move it (perhaps once we clean up all changes from this week)

PR #211, opened 2016-03-08
Motion planning panel

yimingyanged:
Added motion planning panel (under teleop panel).
the new panel includes
- end-effector planning
- walking planning
- fixed-feet collision free motion planning

Currently, walking planner calls the default foot step planner. End-effector and collision free planners call either matlab IKserver or exotica to solve IK and planning problems.

Some of the parameters are not passed to exotica, one still needs to setup the parameters and options via exotica xml config file.

@MarcoCar


patmarion:
cool!  Would love to see some screenshots or video demo :)

yimingyanged:
I am not in the lab right now, I will create some screenshots and videos tomorrow morning.

yimingyanged:
Hi Pat @patmarion ,
Here is the video link [https://youtu.be/RpnrVSp_VJk](https://youtu.be/RpnrVSp_VJk)

There are 3 sub panels under motion planning, 
1) end-pose planning; 
2) walking planning 
3) fixed-feet motion planning. 

The action sequence should be: first plan an end-pose (goal configuration and stance pose), then use walking planner to walk to the stance pose, and finally use fixed-feet motion planning algorithms to generate a collision free trajectory.

- The director needs to be launched with -exo (exotica option) to use the collision-free end-pose planner (iDRM, inverse Dynamic Reachability Map) and Exotica-OMPL motion planning algorithms.

- If run director without -exo, the default algorithms will be used, e.g. IKServer and RRTConnect.

src/python/director/motionplanningpanel.py
```diff
@@ -0,0 +1,391 @@
...
+def _getAction():
+    return app.getToolBarActions()['ActionMotionPlanningPanel']
+
+def init(robotStateModel, robotStateJointController, teleopRobotModel, teleopJointController, debrisPlanner, manipPlanner, affordanceManager, showPlanFunction, hidePlanFunction, footDriver):
+
+    global panel
+    global dock
+
+    panel = MotionPlanningPanel(robotStateModel, robotStateJointController, teleopRobotModel, teleopJointController, debrisPlanner, manipPlanner, affordanceManager, showPlanFunction, hidePlanFunction, footDriver)
+    panel.deactivate()
```
patmarion: is this call to deactivate needed? it would be nice to make sure all the initialization if done in the MotionPlanningPanel constructor

src/python/director/motionplanningpanel.py
```diff
@@ -0,0 +1,391 @@
...
+        
+        
+    def getCurrentWalkingGoal(self):
+        t = self.footDriver.getFeetMidPoint(self.teleopRobotModel)
+        t.PreMultiply()
+        t.Translate(0.0, 0.0, 0.0)
+        t.PostMultiply()
+        return t
+    
+    def onWalkingPlan(self):
```
patmarion: it seems like this function has a lot of copy and pasted code from the onNewWalkingGoal() function in footstepsdriverpanel.py.  Perhaps the common code could be placed into footstepsdriver.py so that it may be reused?

patmarion:
looks great!  I made a few inline comments.  I think this is ready to merge after those are resolved.

mauricefallon:
@yimingyanged : could I get a demo before we merge this? I want to make sure this panel can be shared with Marco and to see how it might be improved to get it running on Val
It seems great though.

This panel is closer to supporting proper collision free motion planning than the telep panel but I see planning a something that is called from within the task sequencing so the task sequencing needs to talk to elements and setting in this panel

yimingyanged:
I was copying stuff from footstepsdriver to make things work without looking into the details. Now I removed the unnecessary codes:
- removed interactive component (since we don't manually set walking goal)
- removed uneven terrain stuff (for now, we only consider flat floor in motion planning)

patmarion:
Looks good.  @mauricefallon merge when ready.

PR #212, opened 2016-03-09
Image Widgets: auto-detect aspect ratio for ImageOverlayManager, make ImageWidget default

iamwolf:
This PR includes two changes to the ImageOverlayManager and ImageWidget:

1. ImageOverlayManager now auto-detects the aspect ratio of the image to be displayed so that there is no more black borders at the top and bottom of the image
2. ImageWidget can be initialised hidden
3. ImageWidget is now the default for the toggle button in the user interface. The rationale behind this is that we often record videos with an image overlay on only to realise afterwards that it's a black spot instead of the video
4. The source of an ImageWidget can be changed without having to re-initialise the object

Pat, there are two remaining issues that I'd like to get your input on:

1. ImageWidget has unused padding left and right of the image, yet I could not find the right vtk command to get rid of it
2. ImageOverlayManager had a better image quality than ImageWidget does

Thanks :)

### DIFF
--- a/src/python/director/cameraview.py
+++ b/src/python/director/cameraview.py
@@ -416,6 +416,9 @@ class ImageWidget(object):
         self.imageName = imageName
         self.flip.SetInput(imageManager.getImage(imageName))
 
+    def setOpacity(self, opacity=1.0):
+        self.imageWidget.GetRepresentation().GetImageProperty().SetOpacity(opacity)
+
     def hide(self):
         self.imageWidget.Off()
 
--- a/src/python/director/cameraview.py
+++ b/src/python/director/cameraview.py
@@ -395,6 +395,9 @@ class ImageWidget(object):
 
         self.imageWidget = vtk.vtkLogoWidget()
         rep = self.imageWidget.GetRepresentation()
+        rep.ProportionalResizeOn()
+        rep.SetPosition(0, 0.7)
+        rep.SetPosition2(0.3, 0.3) # Proportional size
         rep.GetImageProperty().SetOpacity(1.0)
         self.imageWidget.SetInteractor(self.view.renderWindow().GetInteractor())
 
--- a/src/python/director/cameraview.py
+++ b/src/python/director/cameraview.py
@@ -395,23 +395,42 @@ class ImageWidget(object):
         self.initialized = False
 
         self.imageWidget = vtk.vtkLogoWidget()
-        rep = self.imageWidget.GetRepresentation()
-        rep.ProportionalResizeOn()
-        rep.SetPosition(0, 0.7)
-        rep.SetPosition2(0.3, 0.3) # Proportional size
-        rep.GetImageProperty().SetOpacity(1.0)
+        imageRep = self.imageWidget.GetRepresentation()
+        self.imageWidget.ResizableOff()
+        self.imageWidget.SelectableOn()
+
+        imageRep.GetImageProperty().SetOpacity(1.0)
         self.imageWidget.SetInteractor(self.view.renderWindow().GetInteractor())
 
         self.flip = vtk.vtkImageFlip()
         self.flip.SetFilteredAxis(1)
         self.flip.SetInput(imageManager.getImage(imageName))
-        rep.SetImage(self.flip.GetOutput())
+        imageRep.SetImage(self.flip.GetOutput())
 
         self.timerCallback = TimerCallback()
         self.timerCallback.targetFps = 60
         self.timerCallback.callback = self.updateView
         self.timerCallback.start()
 
+        self.initialized = False
+
+    def setWidgetSize(self, desiredWidth=400):
+        viewWidth, viewHeight = self.view.width, self.view.height
+        defaultImageAspectRatio = 4/3.
+        imageWidth, imageHeight = desiredWidth, desiredWidth/defaultImageAspectRatio
+
+        rep = self.imageWidget.GetBorderRepresentation()
+        rep.SetShowBorderToOff()
+        coord = rep.GetPositionCoordinate()
+        coord2 = rep.GetPosition2Coordinate()
+        coord.SetCoordinateSystemToDisplay()
+        coord2.SetCoordinateSystemToDisplay()
+        coord.SetValue(0, viewHeight-imageHeight)
+        coord2.SetValue(imageWidth, imageHeight)
+
+        if self.visible:
+            self.hide()
+            self.show()
 
     def setImage(self, imageName):
         self.imageName = imageName
@@ -425,6 +444,10 @@ class ImageWidget(object):
         self.imageWidget.Off()
 
     def show(self):
+        if not self.initialized:
+            self.setWidgetSize(400)
+            self.initialized = True
+
         self.visible = True
         self.imageWidget.On()
 
--- a/src/python/director/cameraview.py
+++ b/src/python/director/cameraview.py
@@ -428,9 +428,7 @@ class ImageWidget(object):
         coord.SetValue(0, viewHeight-imageHeight)
         coord2.SetValue(imageWidth, imageHeight)
 
-        if self.visible:
-            self.hide()
-            self.show()
+        self.view.render()
 
     def setImage(self, imageName):
         self.imageName = imageName
--- a/src/python/director/cameraview.py
+++ b/src/python/director/cameraview.py
@@ -412,6 +412,10 @@ class ImageWidget(object):
         self.timerCallback.start()
 
 
+    def setImage(self, imageName):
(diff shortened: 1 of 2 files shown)


src/python/director/startup.py
```diff
@@ -1,6 +1,8 @@
 # This script is executed in the main console namespace so
 # that all the variables defined here become console variables.
 
+from __future__ import division
```
patmarion: im a wimp and so i am concerned about adding this to startup.py.  I don't mind using it in modules, but this effects everything run in the python console.  if you think it should be here then maybe i can be convinced.

iamwolf: It turns out this did *not* extend to the console, as much as I'd liked it to as future imports are on a per module basis (cf. [here](https://docs.python.org/2/reference/simple_stmts.html#future)). You can test it by pulling the code, compiling and starting director, firing up the console and typing 4/3, the result is 1. I'm curious to understand the potential impact of this import (I'm happy to change the statements below to explicitly return floats on division if you prefer)

screenshot from 2016-03-09 18 31 06

patmarion:
This code positions the imageWidget in the top left corner and disables border/resizable/movable:


w = imageWidget.imageWidget
rep = w.GetBorderRepresentation()

rep.SetShowBorderToOff()
w.ResizableOff()
w.SelectableOn()

coord = rep.GetPositionCoordinate()
coord2 = rep.GetPosition2Coordinate()

viewWidth, viewHeight = view.width, view.height
imageWidth, imageHeight = 400, 400

coord.SetCoordinateSystemToDisplay()
coord2.SetCoordinateSystemToDisplay()

coord.SetValue(0, viewHeight-imageHeight)
coord2.SetValue(imageWidth, imageHeight)


The trick here is setting the coordinate system to "display", it defaults to "normalized display".  You could get the exact same result if you use the default "normalized display" you just have to multiply by the view width/height.  In "normalized" mode, the coordinate 1.0, 1.0 == viewWidth, viewHeight.

This locks the image widget to be to the top of the view.  If the view height changes, you need to have a callback (catch the resize event) to reposition the image.  Without a callback, you can only lock the image to the lower left corner, since that is the 0,0 origin in vtk. (In Qt, the 0,0 origin is the top left).

Since you will have to use a callback, you could keep the default "normalized display" coordinate system, since you will have to respo

iamwolf:
Thanks Pat for the detailed explanation and example - I'll incorporate it and will ping once it's done

iamwolf:
@patmarion Please take another look - I've incorporated the image resizing and it looks great - thanks so much for the explanation and detailed tutorial :)

src/python/director/startup.py
patmarion: ok, i see in the docs that the import will only be valid for code in startup.py and won't extend to the console, since startup.py is run with an exec statement.

Would you mind removing the import?

iamwolf:
updated, PTAL

patmarion:
The new image widget behavior seems a little buggy to me.  I think it's probably easy to fix.  Have you experienced that?  At least when I try it, the image does not dock to the left edge of the view, the screen does not always re-render when toggling it on/off, and it doesn't reposition if the view size changes.

wxmerkt:
Yeah, unfortunately we have seen some of them. Did you mostly experience that with the Multisense by the way? Docking works fine for 3:2 views but breaks for the 1:1 of the Multisense (a manual imageWidget.resize() or imageWidget.setSize(400,400) resolves it). 
Do you have any idea why it might not re-render?
Would linking to the window size changed signal (to trigger a resize with equal size) help?
Happy to take a closer look and submit a fix.

src/python/director/startup.py
```diff
@@ -685,38 +686,45 @@ def onPlanStatus(msg):
...
     cameraFrustumVisualizer = cameraview.CameraFrustumVisualizer(robotStateModel, cameraview.imageManager, 'CAMERA_LEFT')
 
 class ImageOverlayManager(object):
 
     def __init__(self):
         self.viewName = 'CAMERA_LEFT'
-        #self.viewName = 'KINECT_RGB'
-        self.size = 400
+        self.desiredWidth = 400
+        self.size = [self.desiredWidth, self.desiredWidth/(4/3.)]
```
patmarion: this should be reverted, the aspect ration needs to be defined by the image dimensions after the first image is received by lcm, it should not be hard coded

src/python/director/startup.py
```diff
@@ -685,38 +686,45 @@ def onPlanStatus(msg):
...
         self.position = [0, 0]
         self.usePicker = False
         self.imageView = None
         self.imagePicker = None
         self._prevParent = None
+        self.imageSize = [640, 480]
+        self.imageAspectRatio = 4/3.
```
patmarion: same here

src/python/director/startup.py
```diff
@@ -685,38 +686,45 @@ def onPlanStatus(msg):
...
             return
 
         imageView = cameraview.views[self.viewName]
         self.imageView = imageView
         self._prevParent = imageView.view.parent()
 
+        imageExtent = cameraview.imageManager.images[self.viewName].GetExtent()
+        if imageExtent[1] != -1 and imageExtent[3] != -1:
+            self.imageSize = [imageExtent[1]+1, imageExtent[3]+1]
+            self.imageAspectRatio = self.imageSize[0] / self.imageSize[1]
```
patmarion: oh i see it is set here.  this is not the best place to set it because this code is only called the first time the overlay is shown, which might be before the image exists.

imageAspectRatio should not need to be a stored variable, it's an attribute of the image data

src/python/director/startup.py
```diff
@@ -685,38 +686,45 @@ def onPlanStatus(msg):
...
+            self.size = [self.desiredWidth, self.desiredWidth / self.imageAspectRatio]
+
         imageView.view.hide()
         imageView.view.setParent(view)
-        imageView.view.resize(self.size, self.size)
+        imageView.view.resize(*self.size)
```
patmarion: this is a little hard to read, could just be self.size[0], self.size[1]

src/python/director/startup.py
```diff
@@ -749,9 +757,13 @@ def toggle(self):
...
 
 imageOverlayManager = ImageOverlayManager()
-imageViewHandler = ToggleImageViewHandler(imageOverlayManager)
+imageWidget = cameraview.ImageWidget(cameraview.imageManager, 'CAMERA_LEFT', view, visible=False)
+imageViewHandler = ToggleImageViewHandler(imageWidget)
 showImageOverlay = imageOverlayManager.show
 hideImageOverlay = imageOverlayManager.hide
+showImageWidget = imageWidget.show
+hideImageWidget = imageWidget.hide
+setImageWidgetSource = imageWidget.setImage
```
patmarion: does anyone actually use these functions?  The show/hide functions were originally for convenience before we had a toggle button.  Now that we have a toggle button in the toolbar, we should just delete all these functions

patmarion:
sorry, i know this is an old pull request but i guess i didn't review it so closely at the time.  made some inline comments.  no rush to fix.

src/python/director/cameraview.py
```diff
@@ -385,44 +385,77 @@ def updateView(self):
...
+        coord.SetCoordinateSystemToDisplay()
+        coord2.SetCoordinateSystemToDisplay()
+        coord.SetValue(0, viewHeight-imageHeight)
+        coord2.SetValue(imageWidth, imageHeight)
+
+        self.view.render()
+
+    def setImage(self, imageName):
+        self.imageName = imageName
+        self.flip.SetInput(imageManager.getImage(imageName))
```
patmarion: i think this method setImage(imageName) could be removed, no need to support this api.  is anyone using this method?  or keep it but change the function name to setImageName()

wxmerkt: I'll change it to setImageName (using it to change the underlying image being displayed in different scenarios)

src/python/director/startup.py
wxmerkt: Removed

src/python/director/startup.py
wxmerkt: Addressed, cf. #253

src/python/director/startup.py
wxmerkt: Cf #253: Now in a method (_updateAspectRatio) and called as necessary

src/python/director/startup.py
wxmerkt: Removed, cf. #253

src/python/director/startup.py
wxmerkt: Removed, cf. #253

PR #220, opened 2016-03-14
Updates to footsteps panel in ui.

simalpha:


patmarion:
Looks very nice, great job!  There are a couple things I want to discuss with @rdeits in order to review this, but I may not have time to get to it today.  Is it ok if we keep this one open for a little while?  Do you guys have a workaround that you can use to set the ihmc controller parameters in the meantime?

simalpha:
No problem. I guess we could use my branch for the moment.

src/python/director/footstepsdriver.py
```diff
@@ -20,6 +20,7 @@
 import numpy as np
 from director import drcargs
 import drc as lcmdrc
+import ihmc
```
iamwolf: ``import ihmc as lcmihmc`` would be coherent with our current notation

src/python/director/footstepsdriver.py
```diff
@@ -90,35 +95,47 @@
...
                                  'Map Mode': 0}}
-DEFAULT_STEP_PARAMS['Terrain'] = DEFAULT_STEP_PARAMS['Drake Nominal'].copy()
-DEFAULT_STEP_PARAMS['Terrain'].update({'Drake Min Hold Time': 1.0,
-                                       'Drake Swing Speed': 0.6,
-                                       'Swing Height': 0.05,
+
+DEFAULT_PLANNING_PARAMS['Terrain'] = DEFAULT_PLANNING_PARAMS['Drake Nominal'].copy()
+DEFAULT_PLANNING_PARAMS['Terrain'].update({#'Drake Min Hold Time': 1.0,
+                                       #'Drake Swing Speed': 0.6,
+                                       #'Swing Height': 0.05,
```
patmarion: I'm not sure, but I think it is a bug to not include these parameters in this parameter set.

Thinking about this a little more, I'm not so sure any more that it adds value to separate planning and control parameters.  One could argue that the separation is a detail that should not be exposed to the user.  To the user, they are parameters that affect walking.  In theory, a planner may need to be told about swing speed in order to compute a correct plan.

simalpha:
@patmarion 
Please, refer to this [video](https://www.dropbox.com/s/6yap1pnwj31h6vy/modified_footsteps_panel.mp4?dl=0). It shows what we would like to achieve in the ui. We would like both to separate in two the planning parameters (just one list is too long and does not fit in the window) and to be able to set the ihmc params.

The panel I show you in the video is the result I get with the code on my branch. I understand what are your doubts. I would say that you are correct when you say "I think it is a bug to not include these parameters in this parameter set". Could you modify the ui at my place to achieve similar results with the best strategy?

patmarion:
Thanks for posting the video.

I understand the desire for having a more compact list of properties.  Perhaps there is something we can do in the UI presentation to achieve that without splitting the parameters into separate groups at the code layer.  The problem with this pull request is that by separating parameters into groups it is breaking the "presets" feature which is something we use for Atlas and you may want to use in your lab too.

And just to repeat a comment I made in the other PR, I'm not convinced (but maybe I could become convinced) that splitting parameters into planning and control groups is a good idea, to the user (and task planner) they are all parameters that affect walking behavior.  The fact that some parameters are only read by the controller seems like an implementation detail.

I'll think about UI designs that could achieve our goals.  I hope in the meantime you guys can find a workaround so you can be productive, I don't want to block you, but as it stands this PR is breaking some functionality so I don't want to merge it as is.  Perhaps the simplest modification to move forward is just adding the IHMC parameters to the list, making the UI even longer, but hopefully soon we can find a satisfying redesign to resolve it.

PR #221, opened 2016-03-14
Add vrml support

patmarion:


patmarion:
todo- id like to add a test that loads all the support file formats using data files from drc-testing-data.

support for vrml was motivated so that director could open vrml files generated by HACD-

http://kmamou.blogspot.ca/2014/11/v-hacd-v20-is-here.html

mauricefallon:
Looks cool! quality of mesh convex decomposition seems underdeveloped

PR #226, opened 2016-03-18
Drake visualizer triangle mesh

patmarion:


patmarion:
This adds support for drawing meshes in drake-visualizer by passing the mesh data arrays inside the lcm message, rather than passing a mesh filename.

Normally, the mesh_filename is passed with string_data.  With this pull request, if string_data is empty, then drake-visualizer will assume that float_data contains mesh data arrays.  This is a bit hacky, but it is simple enough.  The float_data should encode the vertices and list of triangle vertex indices as a flat array: 

    [num_points, num_tris, xyz points, vert indices]

patmarion:
I am going to merge this since it is compatible with existing drake draw messages, it does not interfere.  In the future, we may want to develop richer support for mesh array data, but this is useful for now.

amcastro-tri:
@patmarion, would it be possible to use a similar interface to addres  RobotLocomotion/drake/issues/1908?

patmarion:
I'll post a response over there.

PR #229, opened 2016-03-21
Neck Control Updates

wxmerkt:
- Removed neck pitch from Multisense Panel
- Added the neck pitch setting for Atlas to the Teleop Panel as well, changed Valkyrie's neck command channel
- Changed message type for DESIRED_NECK_ANGLES to joint_angles_t

### DIFF
--- a/src/python/director/teleoppanel.py
+++ b/src/python/director/teleoppanel.py
@@ -1135,14 +1135,15 @@ class JointTeleopPanel(object):
             if groupName == 'Neck':
                 def onSendNeckJointPositionGoal():
                     if 'neck_ay' in joints:  # Atlas
-                        neckPitchDegrees = self.getJointValue(self.toJointIndex('neck_ay'))
+                        neckPitch = self.getJointValue(self.toJointIndex('neck_ay'))
                         msg = lcmdrc.neck_pitch_t()
                         msg.utime = 0
-                        msg.pitch = neckPitchDegrees
+                        msg.pitch = neckPitch
 
                         lcmUtils.publish('DESIRED_NECK_PITCH', msg)
                     elif 'lowerNeckPitch' in joints:  # Valkyrie
                         msg = lcmbotcore.joint_angles_t()
+                        msg.utime = 0
                         msg.num_joints = len(joints)
                         msg.joint_name = joints
                         msg.joint_position = [0] * len(joints)
--- a/src/python/director/teleoppanel.py
+++ b/src/python/director/teleoppanel.py
@@ -12,6 +12,7 @@ from director import vtkAll as vtk
 from director import drcargs
 from director import affordanceurdf
 from director.roboturdf import HandFactory
+from director.utime import getUtime
 from director import lcmUtils
 import drc as lcmdrc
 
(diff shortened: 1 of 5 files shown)


src/python/director/teleoppanel.py
```diff
@@ -1134,17 +1134,23 @@ def buildTabWidget(self, jointGroups):
...
-                    msg.joint_position = [0] * len(joints)
-                    for i, jointName in enumerate(joints):
-                        jointIndex = self.toJointIndex(jointName)
-                        msg.joint_position[i] = self.getJointValue(jointIndex)
-
-                    lcmUtils.publish("JOINT_POSITION_GOAL", msg)
+                    if 'neck_ay' in joints:  # Atlas
+                        neckPitchDegrees = self.getJointValue(self.toJointIndex('neck_ay'))
+                        msg = lcmdrc.neck_pitch_t()
+                        msg.utime = 0
```
patmarion: i know this field wasn't set in the old code, but it should be set

src/python/director/teleoppanel.py
```diff
@@ -1134,17 +1134,23 @@ def buildTabWidget(self, jointGroups):
...
-                    msg.joint_velocity = [0] * len(joints)
-                    msg.joint_effort = [0] * len(joints)
-                    msg.joint_position = [0] * len(joints)
-                    for i, jointName in enumerate(joints):
-                        jointIndex = self.toJointIndex(jointName)
-                        msg.joint_position[i] = self.getJointValue(jointIndex)
-
-                    lcmUtils.publish("JOINT_POSITION_GOAL", msg)
+                    if 'neck_ay' in joints:  # Atlas
+                        neckPitchDegrees = self.getJointValue(self.toJointIndex('neck_ay'))
```
patmarion: variable name is wrong, it is not degrees here, which is ok, the message takes radians

src/python/director/teleoppanel.py
```diff
@@ -1134,17 +1134,23 @@ def buildTabWidget(self, jointGroups):
...
+                        msg = lcmdrc.neck_pitch_t()
+                        msg.utime = 0
+                        msg.pitch = neckPitchDegrees
+
+                        lcmUtils.publish('DESIRED_NECK_PITCH', msg)
+                    elif 'lowerNeckPitch' in joints:  # Valkyrie
+                        msg = lcmbotcore.joint_angles_t()
+                        msg.num_joints = len(joints)
+                        msg.joint_name = joints
+                        msg.joint_position = [0] * len(joints)
```
patmarion: i think this message has a utime field that you could fill in

patmarion:
should we consolidate so that the atlas driver can receive joint_angles_t on the channel DESIRED_NECK_ANGLES?

Also, maybe you haven't used the feature before, but there is a keyboard shortcut to set neck presets, such as looking forward and looking down.  Those shortcuts are activated by pressing N+0, N+1, etc, and handled by the NeckDriver class in perception.py.  That class isn't very fancy, but if it's useful on Valkyrie, we could try to reuse it.

patmarion:
for utime field, you can do this:

from director.utime import getUtime


msg.utime = getUtime()

wxmerkt:
Thanks, I've just added the utime setting this as you were commenting :)
Thanks for the review - I will also take a look at the Atlas driver and panels to change it there as well - I presume there are no backwards compatibility issues with changing the message type and channel?

mauricefallon:
> should we consolidate so that the atlas driver can receive joint_angles_t on the channel 
> DESIRED_NECK_ANGLES?

+1

wxmerkt:
Updated this PR; now dependent on https://github.com/openhumanoids/atlas-drivers/pull/4.

patmarion:
looks good to me

patmarion:
please go ahead and merge when you're ready

PR #232, opened 2016-03-23
Updates to footsteps panel in ui [replacement]

mauricefallon:
I added to Simona's branch by:
- using only a single options panel
- splitting the first 15 parameters into one pane and the other 14 into a second pane.

It appears outwardly the same as for https://github.com/RobotLocomotion/director/pull/220


mauricefallon:
Short video demoing:

https://www.dropbox.com/s/8mcpwdy7j21irwt/2016-03-multi-panels.mkv?dl=0

patmarion:
+1

I like the new feature you added to the property panel, this seems really useful for displaying subsets.

mauricefallon:
Thanks. A next step would be to reduce the size of the plan playback panel. We have a lot of 1080p monitors.

Of course less properties would solve this problem too ;)

PR #240, opened 2016-03-28
add an interactive pydrake ik test - work in progress

patmarion:


patmarion:
requires a few commits from my drake branch here:

https://github.com/patmarion/drake/commits/update-swig-bindings

patmarion:
This test script runs interactive ik with fixed foot constraints, quasi static  constraint, and left hand 6dof constraint.  Ik is called through the pydrake swig bindings.  It's much faster than serializing constraints and calling matlab :+1: 

Run with:

    directorPython src/python/tests/testPyDrakeIk.py

The test also works with -v5 argument to run with Atlas.

The test requires a few drake commits from my branch here:

https://github.com/patmarion/drake/commits/update-swig-bindings

rdeits:
woohoo!

mauricefallon:
woohoo +1 

@patmarion : do you want to make the running on this (you might prefer that)? We are keen to help and test

patmarion:
Yes, that works for me, but I can't propose a timeline yet I have a few other things I'm working on.

PR #241, opened 2016-03-29
use a minimum bounding box algorithm to recover the frame of a table

mauricefallon:
- stops using getOrientedBoundingBox (vtkAnnotateOBBs). reuses code from continuous walking block segmentation. in particular gives more reliable standing frame. still needs some improvement but works better than previous approach

edgePoint is no longer required for segmentTableEdge, but haven't changed it in argument.

Before (can actually be much worse):
before

After:
after

### DIFF
--- a/src/python/director/segmentation.py
+++ b/src/python/director/segmentation.py
@@ -26,6 +26,9 @@ from director.fieldcontainer import FieldContainer
 from director.segmentationroutines import *
 from director import cameraview
 
+from thirdparty import qhull_2d
+from thirdparty import min_bounding_rect
+
 import numpy as np
 import vtkNumpy
 from debugVis import DebugData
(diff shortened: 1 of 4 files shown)


patmarion:
by the way, I also re-used this code for a box picking demo I worked on.  I have not had a chance to clean up my code and commit it yet.  In my code, I removed the findFarRightCorner() function, I think it is overly complex.  For the DRC finals terrain task I developed some general purpose re-orientation code.  If I get around to it, when I clean up my code, I might incorporate it so that we can remove the findFarRightCorner dependency.

src/python/director/segmentation.py
```diff
@@ -2529,47 +2532,33 @@ def segmentTableEdge(polyData, searchPoint, edgePoint):
...
-
-    edgeCenter = origin - 0.5 * axes[0]*edgeLengths[0] + 0.5*axes[2]*edgeLengths[2]
-    edgeLeft = edgeCenter + 0.5 * axes[1]*edgeLengths[1]
-    edgeRight = edgeCenter - 0.5 * axes[1]*edgeLengths[1]
-
-    t = getTransformFromAxes(axes[0], axes[1], axes[2])
-    t.PostMultiply()
-    t.Translate(edgeRight)
+    # TODO: replace this frame with view frame. (Currently viewframe is inverted on Valkyrie)
+    viewFrame = getLinkFrame("pelvis")# SegmentationContext.getGlobalInstance().getViewFrame()
```
patmarion: In the other comment I made, about the re-orientation code used for the drc finals terrain task, we used the stance frame as the reference frame.  The stance frame is usually very similar to the pelvis frame, though at a different z height obviously.  We have other tasks, too, where the stance frame is used to define "robot forward".  It's just slightly more generalized than using a named link frame such as pelvis, torso, or head.

mauricefallon:
You're right - the look direction is enough and the head inversion shouldnt be a blocking issue. I'll clean this up further shortly.

Issue #235, opened 2016-03-24
Fix failing test


patmarion:
The testPlanConstraints.py test has been failing for a while.  I looked and it has started failing due to
commit 9cf03acfd4e152a562e76a0cfe28d796c0a8fbbf which was part of the motion planning panel pull request.  That commit made a change to ikplanner.py and the commit message says that it was intended to fix a bug.

@yimingyanged can you explain what the bug was?  I think I'd like to revert that commit, but I don't know how it will affect the bug you were concerned with.

patmarion:
@yimingyanged I have opened a pull request to revert your commit.  Please let me know if you have a test or method to reproduce the bug you were fixing with that commit, because I'm not sure what it is about.

PR #242, opened 2016-03-29
Fix ik matlab communication

yimingyanged:
@patmarion  Sorry for the late reply. The issue in #235 #236 is because pushToMatlab is used in a conflict way. It is used to indicate both 'push to matlab or not' and 'use matlab or exotica to solve IK'. I added a new variable to select the planning mode (for now, drake or exotica). The pushToMatlab var is now only used to indicate whether push things to matlab.

### DIFF
--- a/src/python/director/startup.py
+++ b/src/python/director/startup.py
@@ -1086,12 +1086,6 @@ if 'useKuka' in drcargs.getDirectorConfig()['userConfig']:
     #ikPlanner.fixedBaseArm = True
     #showImageOverlay()
 
-
-if 'exo' in drcargs.args():
-    if (drcargs.args().exo):
-        ikPlanner.pushToMatlab = False
-
-
 def roomMap():
     mappingPanel.onStartMappingButton()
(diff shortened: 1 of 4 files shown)


src/python/director/startup.py
```diff
@@ -1086,12 +1086,6 @@ def stop(self):
     #ikPlanner.fixedBaseArm = True
     #showImageOverlay()
 
-
-if 'exo' in drcargs.args():
```
wxmerkt: I believe this also requires a change to https://github.com/openhumanoids/director/blob/fix-ik-matlab-communication/src/python/director/drcargs.py#L77 or replace the functionality here to set ``self.ikPlanner.planningMode = 'exotica'`` if exo is in drcargs.args()

patmarion:
could you change this so that only the `addPose` function checks the  `self.pushToMatlab` variable?  The other functions should not use `pushToMatlab`, the variable was only ever meant to switch behavior or the `addPose` function.

yimingyanged:
@patmarion I have removed pushToMatlab in runIK and runIKTraj

patmarion:
thanks!

PR #243, opened 2016-03-31
load walking environment in UoE

mauricefallon:


patmarion:
If you'd like to do a skype call, I can demo the terrain code developed for walking the drc terrain.  Maybe it would be useful.

mauricefallon:
this is something much more basic. its just an environment for testing in sim

PR #245, opened 2016-04-01
Updated motion planning panel

yimingyanged:
- Added bi-hands reaching option
- Added walking plan fake execute option
- Some UI updates

### DIFF
--- a/src/python/director/motionplanningpanel.py
+++ b/src/python/director/motionplanningpanel.py
@@ -82,7 +82,7 @@ class MotionPlanningPanel(object):
         # Foot step planning
         self.placer = None
         self.ui.walkingPlanButton.connect('clicked()', self.onWalkingPlan)
-        
+        self.ui.walkingFakeExeButton.connect('clicked()', self.onWalkingFakeExecute)
         # Motion Planning
         self.ui.motionPlanButton.connect('clicked()', self.onMotionPlan)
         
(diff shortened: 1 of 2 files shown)


src/python/director/motionplanningpanel.py
```diff
@@ -82,7 +82,7 @@ def __init__(self, robotStateModel, robotStateJointController, teleopRobotModel,
         # Foot step planning
         self.placer = None
         self.ui.walkingPlanButton.connect('clicked()', self.onWalkingPlan)
-        
+        self.ui.walkingFakeExeButton.connect('clicked()', self.onWalkingFakeExecute)
```
mauricefallon: what is "fake exe"

mauricefallon:
seems reasonable but we use "teleport" to mean "moving robot model in the UI by not in reality"

src/python/director/motionplanningpanel.py
yimingyanged: 'fake exe' simply move the robot to the desired location without actual 'walking'. It provides a way to evaluate and test final pose planning and motion planning in a much simpler manner

yimingyanged: It is similar to MoveIt's fake execution (where I got the name from)

mauricefallon: i understand, please change to teleportRobotToStanceFrame can see if you can make it identical to the version in tabledemo.py and valvedemo.py

where we have common functions like this we often move them to robottasks.py

PR #247, opened 2016-04-04
Add findOutputDevice to midi.py

wxmerkt:
Mirroring what's there for findInputDevice

By the way, was there a specific reason not to use the python-midi package that can be installed via pip?


patmarion:
I just gave a quick look at the pip package, are you sure it reads from midi devices?  It looks like it might only do midi files.

PR #248, opened 2016-04-05
Updated swing and transfer time defaults.

simalpha:
I updated default values for swing and transfer time according to results of walking evaluation ([Valkyrie Walking Evaluation](https://github.com/openhumanoids/wiki/wiki/Valkyrie-Walking-Evaluation)).


mauricefallon:
The scientific method!

https://en.wikipedia.org/wiki/Scientific_method

PR #249, opened 2016-04-06
Added a fix to table segmentation 

raluca-scona:
This uses the viewFrame position and the yaw from viewDirection as the head frame


mauricefallon:
merging but would like to make this function cleaner. need to collect test data of different tables at different slants.

PR #252, opened 2016-04-14
fixes for the image widget

patmarion:

PR #253, opened 2016-04-14
ImageWidget: Follow up to #212, #252

wxmerkt:
Thanks for the comments and fix in #212 @patmarion - this PR addresses these comments and cleans up the aspect ratio.


patmarion:
thanks!

PR #254, opened 2016-04-14
mark double click event as consumed after toggling footstep widget

patmarion:
i broke this when splitting view behaviors into robot view behaviors


patmarion:
the double click was activating the footstep widget, but then the double click event triggered a second time in the view behaviors base class, toggling the frame off.  oops.

PR #255, opened 2016-04-20
[minor] Turn off foot contact state visualisation

wxmerkt:
It's not working reliably at the moment as tared values are not exposed. Will reactivate once we can reliably show contact state information


patmarion:
Maybe you could add a toggle option to the Tools menu, as we have for several other features, and set it to be off by default?

wxmerkt:
I was just about to write that this was rather stupid of me to merge this as is since it's the same global variable for Atlas and Valkyrie. I will put it on my todo to add logic to the Valkyrie Driver / when run with the modules/ohdirector/runstartup.py to turn it selectively on/off. Please revert this PR as necessary

PR #256, opened 2016-04-21
Make "all" the default target in distro/pods/drake-distro.

david-german-tri:
This contributes to https://github.com/RobotLocomotion/drake/issues/2136.  

I'm sending you this PR because I can't think of a reason you would want `VERBOSE=1 make` to fail.  If there is a reason, we can instead change the Drake build to build the `all` target explicitly.

### DIFF
--- a/distro/pods/drake-distro/Makefile
+++ b/distro/pods/drake-distro/Makefile
@@ -1,10 +1,5 @@
 # Default makefile distributed with pods version: 12.01.11
 
-
-# Default to a less-verbose build.  If you want all the gory compiler output,
-# run "make VERBOSE=1"
-$(VERBOSE).SILENT:
-
 # Figure out where to build the software.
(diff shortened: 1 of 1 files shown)


patmarion:
I think this is a standard pod makefile, are they all broken this way?

In the future we can have drake build director as a standard CMake project and retire this makefile.

david-german-tri:
I guess they are!  That's fun.

Agreed, the day we build director as CMake will be a happy one.  I think @mwoehlke-kitware is working towards that.

patmarion:
Looking at other pod makefiles as an example, adding this line to the top of the makefile would make it match the others:

    default_target: all

I'm not sure why it is missing from the director makefiles.  It appears that your patch has the same effect though.  The fix should be applied to both makefiles:

    distro/pods/drc/Makefile
    distro/pods/drake-distro/Makefile

PR #260, opened 2016-04-28
updates for valkyrie streaming

manuelli:


patmarion:
@wxmerkt do you guys use testAtlasCommand.py for the streaming interface?  We plan to develop this script further to feed desired positions to the QP controller, which we are developing under a new mode (activated by command line argument) but I want to know whether we can freely make mods here without interfering with your use cases.

wxmerkt:
We are using the script occasionally, mostly the ``--robot`` option for the command streaming, or the ``--base`` mode for joint position goal streaming to the ``--robot``. We are considering to use it in the future as far as I know (I'll read the PRs/follow what you'll amend).

PR #268, opened 2016-05-09
add utility scripts to test controller using streaming interface

manuelli:


mauricefallon:
@manuelli : I think this is unneeded. Can you close if so?

manuelli:
Yes this is old. Don't need this functionality now but if I do I'll put it into mitdirector directory.

PR #271, opened 2016-05-24
valkyrie updates

manuelli:
- [ ] 

1. uses new version of qp_controller_input message with splines
2. adds functionality to the atlas/valkyrie control panel

### DIFF
--- a/src/python/director/atlasdriverpanel.py
+++ b/src/python/director/atlasdriverpanel.py
@@ -2,6 +2,8 @@ import PythonQt
 from PythonQt import QtCore, QtGui, QtUiTools
 from director import applogic as app
 from director.timercallback import TimerCallback
+from director import drcargs
+
 from functools import partial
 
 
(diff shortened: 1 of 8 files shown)


src/python/director/lcmUtils.py
```diff
@@ -11,6 +11,8 @@
 import sys
 import re
 
+import drake
```
patmarion: Hello from Big Bear Lake!

Don't add dependency to this file.  There's another lcm message util file that has DRC message dependency, add there.

mauricefallon: haha! Code reviews from on the PCT!

mauricefallon:
Hi @manuelli ,

@patmarion set up the following for us (Edinburgh) so as to have a working environment for Director code within the OH repo (so as to avoid having to juggle 2-3 repos).

By launching the director as follows:
director -val2 --startup $DRC_BASE/software/modules/ohdirector/runstartup.py
The runstartup.py script is run on startup and creates any other classes or modules we are developing.

For example it disables/hides the Atlas panel and creates a Valkyrie panel and some new demo panels.

The big idea is to avoid pushing minor modules and mods into director. I would recommend you making something similar for MIT's operation of Val.

src/python/director/atlasdriverpanel.py
```diff
@@ -27,48 +29,84 @@ def __init__(self, driver):
         self.driver = driver
 
         loader = QtUiTools.QUiLoader()
+
+        # load Atlas by default
         uifile = QtCore.QFile(':/ui/ddAtlasDriverPanel.ui')
+
```
mauricefallon: the runstartup.py arrangement would allow us to avoid making this kind of logical fixes to switch between Atlas and Val

src/python/director/lcmUtils.py
manuelli: ok I'll push these featuers out into the other file.

manuelli:
Thanks for the tip @mauricefallon. It's been a pain having to be doing these minor updates to director all the time. I will break this out into a similar ohdirector type structure for MIT.

manuelli:
I moved this functionality over to the openhumanoids repo like @mauricefallon suggested. Now resides in software/modules/mitdirector

PR #275, opened 2016-06-01
Use bot_core_lcmtypes to provide drake-visualizer interface types

### LINKED ISSUE #273: Building drake-visualizer without Drake
I've been using the drake-visualizer app a lot over the past few weeks from outside the Drake ecosystem (specifically, from https://github.com/rdeits/DrakeVisualizer.jl ), and I think other people might find that to be very useful as well. I think it would be nice to be able to build drake-visualizer without installing Drake itself. As far as I can tell, the only thing stopping us from doing that is the fact that the LCM type definitions for `lcmt_viewer_load_robot` and friends live in Drake. Is that correct? 

If so, I think it would be reasonable to duplicate the message types here so we can build the whole app without Drake.

I'm happy to do the implementation myself. What do y'all think?

rdeits:
This also includes changes from #272 since they end up touching some of the same lines. 

The big exciting feature is that we can now build a fully-functional `drake-visualizer` app without the Drake dependency. 

This adds `bot_core_lcmtypes` as an external to the superbuild, and uses it to provide the types needed by drakevisualizer. If bot_core_lcmtypes doesn't contain the appropriate types, it falls back to extracting them from Drake and throws a deprecation warning. 

With this change, Director no longer depends on any of Drake's python LCM bindings (except in backwards-compatibility mode). 

Closes #273


mauricefallon:
This seems to add the initial ability for a 3rd party to transmit the status of their own kinematic system over lcm and have it renderer by the visualise - which would be cool.

do you have a simple demo script you could add which creates a system and demonstrate this? E.g. adding a python script to director/src/python/tests

rdeits:
Yes! That's exactly the idea. I'm hoping to use this with https://github.com/rdeits/DrakeVisualizer.jl which provides a Julia interface to the visualizer. 

Adding a demo script is a good idea. I'll do that.

src/python/director/drakevisualizer.py
```diff
@@ -289,12 +298,12 @@ def __init__(self, view):
         self.view = view
         self.robots = {}
         self.linkWarnings = set()
-        self.sendStatusMessage('loaded')
         self.enable()
+        self.sendStatusMessage('loaded')
```
rdeits: I switched the order of these lines so that the status message is not sent until *after* the LCM subscriptions are created. Otherwise it's hard to tell when the viewer is ready to start receiving messages.

rdeits:
Ok, script added and commented.

wxmerkt:
LGTM (test has been added and passes), can we merge this @mauricefallon and @rdeits?

mauricefallon:
I don't think we currently use the tip of bot_core_lcmtypes in oh-distro because the changes that TRI added to the CMake didn't work for us.

So I don't think the drake types are in the currently used revision.

mauricefallon:
PS: it would seem that TRI are no longer using bot_core_lcmtypes as they haven't been actively trying to push the open PR in either

wxmerkt:
Drake uses a [fork of bot_core_lcmtypes](https://github.com/mwoehlke-kitware/bot_core_lcmtypes/tree/0434090f6f54228c35514f7c9216e3bf1e003ea3).

Good point regarding the bot_core_lcmtypes, what should we do about this PR then to avoid capability divergence? I feel the feature presented here would be great.

PR #276, opened 2016-06-01
Fix OSX build by switching homebrew taps for vtk5

rdeits:
homebrew/science doesn't include vtk5 anymore, so we have to work around it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -76,7 +76,6 @@ The dependencies can be installed on Mac using `Homebrew <http://brew.sh/>`_:
 
     brew tap homebrew/science
     brew tap homebrew/python
-    brew tap robotlocomotion/director
     brew install cmake
     brew install python numpy
     brew install vtk5 --with-qt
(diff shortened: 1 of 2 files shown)


rdeits:
Sweet. This works. I had to update the bottle to internally point to my vtk5 tap, but it was otherwise pretty easy.

rdeits:
Ok, I think this is ready to go. We may want to change the tap again once https://github.com/RobotLocomotion/homebrew-director comes online. 

Can I get a review?

mauricefallon:
i can't review this as I don't use OSX

wxmerkt:
LGTM. Thanks for fixing this! Would you mind updating the README as well? :)

wxmerkt:
Two more things:
- RL/homebrew-director is now online - should we change this here to use it?
- [for future] The vtk bottle should perhaps be put on a server different from Dropbox. Dropbox blocks links if there's too much traffic (unlikely), sometimes has random timeouts, and in general slows the connection down for public links

rdeits:
Ok, I've switched to robotlocomotion/director and updated the readme. And agreed that dropbox is not a great solution for the future :)

PR #279, opened 2016-06-05
update readme for new vtk5 tap

rdeits:
Note that I'm using the RobotLocomotion/director tap for the Readme (since that'll be the right thing going forward), but I'm using rdeits/director within Travis. That's because I've got exactly the version of the VTK5 recipe that matches the bottled version, so it will keep Travis happy until we make a new bottle.


rdeits:
I also fixed the dropbox bottle, so the OSX build actually passes now.

PR #281, opened 2016-06-22
adding collections renderer

mauricefallon:
This adds a renderer which can draw a variety of point clouds, primitive shapes and transform triads using messages transmitted from an external application.  

Its actually a very mature interface that was well tested in the old GTK lcm viewer. Its similar in many ways to Markers in ROS:
http://wiki.ros.org/rviz/DisplayTypes/Marker

Its supports:
- individual or lists of triads, points, simple polygons, triangles etc.
- addition to or replacement of an entire collection e.g. adding a new triad to an existing collection of triads
- nesting of point clouds relative to transforms. This is useful for mapping applications as point clouds can be moved by changing the transform e.g. moving the frame of a LIDAR sensor to move the point cloud
- remote reset of the collection list
- vary the size and alpha of poses and objects
- dynamically populate the list of objects in the scene graph

Here are two rough videos. I'll submit a tutorial in time:
- [a variety of features](https://www.dropbox.com/s/k019o0pfc54w3qk/collection-demo1.m4v?dl=0)
- [lists of triads being added to](https://www.dropbox.com/s/apt0omfcgumnezf/collection-demo2.m4v?dl=0)
- [triads colored or typical rgb](https://www.dropbox.com/s/a5y5t9eln0pji9p/collection-demo3.m4v?dl=0)

Note: currently the collections lcmtypes are in a pod within pronto, but they could easily be ported to bot_core or an independent repo. collections is only built with the director application at present


mauricefallon:
The original GTK rendering code is here and be melded for easy comparison. Changes are very minimal:
https://github.com/openhumanoids/pronto/blob/master/visualization/src/collections_renderer/collections_renderer.cpp

PR #292, opened 2016-07-15
Fix widget footstep goal computation

raluca-scona:
The translate does not seep to work because selectedGroundPoint and footFrame are in different coordinate frames.


mauricefallon:
seems  to be working, thanks!

PR #294, opened 2016-07-26
rm dead mapping and table demos

mauricefallon:
removes:
- unused mapping panel
- deprecated mapping panel for kuka arm (used in @wxmerkt 's MSc)
- deprecated planning panel, being ported into oh-distro proper


wxmerkt:
I'm pretty sure that the mapping and table demo still work. I'll add them as a Kuka-only module (i.e. only loaded when useKuka is true) to oh-distro tomorrow. Otherwise LGTM

PR #295, opened 2016-07-27
Clamp planning start pose to joint limits

wxmerkt:
Follow up to our lunch discussion with @mauricefallon @yimingyanged @VladimirIvan 
1. Adds ``planningUtils`` in order to assemble methods that are commonly used in motion planning and currently fragmented across multiple files
2. ``planningUtils`` provides a ``getPlanningStartPose()`` method that by default clamps the ``startPose`` of planning problems to the joint limits - thus averting invalid motion plans (that e.g. trip up IHMC's controller)
3. Replaced all uses of manually getting the poses in the teleop and motion planning panels with the new ``getPlanningStartPose`` method

@mauricefallon @yimingyanged would you please have a look over this?

cc: @MarcoCar FYI

**ToDo:**

- [ ] Use ``planningUtils`` across all demos
- [x] remove print output in PlannerPublisher.updateJointLimits()
- [ ] deprecate extending joint limits/leave them untouched for both Matlab as well as EXOTica planners
- [x] make decision flag a property of the class, so that it can be changed globally
- [x] have this property set by a high level variable from startup.py (and in turn from director_config.json)
  - this would mean extending the field 'disableComponents' which can switch off components. There should be both a disableComponents list and an enableComponents list

### DIFF
--- a/src/python/director/planningutils.py
+++ b/src/python/director/planningutils.py
@@ -14,10 +14,13 @@ class PlanningUtils(object):
         self.jointLimitsLower = np.array([self.robotStateModel.model.getJointLimits(jointName)[0] for jointName in robotstate.getDrakePoseJointNames()])
         self.jointLimitsUpper = np.array([self.robotStateModel.model.getJointLimits(jointName)[1] for jointName in robotstate.getDrakePoseJointNames()])
 
-    def getPlanningStartPose(self, clampToJointLimits=True):
+        # Activate/Deactivate joint limit clamper
+        self.clampToJointLimits = False
+
+    def getPlanningStartPose(self):
         startPose = np.array(self.robotStateJointController.q)
 
-        if clampToJointLimits:
+        if self.clampToJointLimits:
             startPose = np.clip(startPose, self.jointLimitsLower, self.jointLimitsUpper)
         
         return startPose
--- a/src/python/director/planningutils.py
+++ b/src/python/director/planningutils.py
@@ -1,0 +1,23 @@
+'''
+PlanningUtils is a centralised helper to provide functions commonly used by 
+motion planners
+'''
+import numpy as np
+from director import robotstate
+
+class PlanningUtils(object):
+    def __init__(self, robotStateModel, robotStateJointController):
+        self.robotStateModel = robotStateModel
+        self.robotStateJointController = robotStateJointController
+
+        # Get joint limits
+        self.jointLimitsLower = np.array([self.robotStateModel.model.getJointLimits(jointName)[0] for jointName in robotstate.getDrakePoseJointNames()])
+        self.jointLimitsUpper = np.array([self.robotStateModel.model.getJointLimits(jointName)[1] for jointName in robotstate.getDrakePoseJointNames()])
+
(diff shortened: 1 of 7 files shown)


src/python/director/planningutils.py
```diff
@@ -0,0 +1,23 @@
...
+class PlanningUtils(object):
+    def __init__(self, robotStateModel, robotStateJointController):
+        self.robotStateModel = robotStateModel
+        self.robotStateJointController = robotStateJointController
+
+        # Get joint limits
+        self.jointLimitsLower = np.array([self.robotStateModel.model.getJointLimits(jointName)[0] for jointName in robotstate.getDrakePoseJointNames()])
+        self.jointLimitsUpper = np.array([self.robotStateModel.model.getJointLimits(jointName)[1] for jointName in robotstate.getDrakePoseJointNames()])
+
+    def getPlanningStartPose(self, clampToJointLimits=True):
```
mauricefallon: what about making clampToJointLimits a variable of the class so that all plans are clamped or not?

otherwise there wouldn't be an easy way of turning this clamp on/off in current code

mauricefallon:
By the way what happens currently?

From what I see, when extending a limit the jointLimitChecker (in extendJointLimitsAsExceeded):
- makes a call to ikServer.updateJointLimits() which writes the limits to matlab
- makes a call to plannerPublisher.updateJointLimits() which I think finds its way to Exotica/LCM

Is that right? The intention would be to ignore this system and to leave the stated limits unchanged - and therefore create no inconsistent set of limits?

@VladimirIvan : Is Exotica sent the set of joint limits in each plan request? Its seems so in PlannerPublisher.setupMessage() because a cache is kept in self.jointLimits. (The Matlab update is just local)

wxmerkt:
Thanks for the feedback, I'll address them Monday. I've updated the to do above accordingly

mauricefallon:
Great, to pull that in fully, we create a variable in the main startup.py (say useLimitJointsSentToPlanner)
In that way we could turn it on and off in the director_config.json - for different robots.

I've just added it to the TODO list

wxmerkt:
@mauricefallon I've addressed the comments, PTAL.

Also, regarding
> deprecate extending joint limits/leave them untouched for both Matlab as well as EXOTica planners

I'm happy to do this - this changes behaviour that's been added back in 2014 though - should we check with Pat whether they are using this feature or would like to also add ``useLimitJointsSentToPlanner`` in their director_config under ``enableComponents``? Otherwise, I'd remove the entire JointLimitChecker or just the updateJointLimits part.

Is it okay to leave the old demo clean-up/update for follow-on PRs? I'll create an issue to track what needs to be done

mauricefallon:
I agree with both of these points.

I like how you've implemented this. My main worry is about sending plan requests when the core joints have been clipped by more than 2 degrees. 
- What kind of safety could we have to avoid planning in that situation?
- We should also add printf when making a heavy clip so we can see that it occurred when playing back a log

Let's leave the old method (red pop-up) in place for now and start using this. Mostly, I don't want the confusion of two different approaches.

I'll accept this and these two issues can be added in testing

PR #297, opened 2016-08-08
Fix unit tests

wxmerkt:
Now that a build server is back up, it became clear that https://github.com/RobotLocomotion/director/pull/295 broke some of our unit tests, cf. [build server failures](http://kobol.csail.mit.edu/cd/viewTest.php?onlyfailed&buildid=23686).

This PR fixes some of them. Additionally, the catkin workspace is now being built on the server a well which should take care of a further few.


mauricefallon:
Awesome @wxmerkt , many thanks!

BTW, test_FPP (from Marco) probably fails because the input data is missing - a 100MB file
@MarcoCar : what do you want to do about this?

MarcoCar:
I created a pull request https://github.com/openhumanoids/oh-distro-private/pull/2529 that should fix that. I will remove that code once I've integrated what I need in exotica.

PR #299, opened 2016-08-18
Support August Drake

gizatt:
API and type tweaks so that Director builds against Drake from early August 2016.


patmarion:
closing pull request until we're actually ready to update director to drake master (which requires buy in from oh-distro), but this patch is very helpful, thanks!

PR #301, opened 2016-09-15
Cleanup some unused code/imports

sammy-tri:
Remove some unneeded imports and unused code.

Background:  I've been working on trying to get basic IK/teleop functionality working in Director using just the Drake tree and with the Python interface to Drake (instead of MATLAB).  This is some preliminary cleanup work which doesn't introduce any functional changes yet.

### DIFF
--- a/src/python/director/ikconstraintencoder.py
+++ b/src/python/director/ikconstraintencoder.py
@@ -3,12 +3,7 @@ import director.thirdparty.numpyjsoncoder as nje
 from collections import OrderedDict
 from director import fieldcontainer
 from director import transformUtils
-from director import lcmUtils
-from director.utime import getUtime
 
-import drc as lcmdrc
-
-import pprint
 import json
 
 class ConstraintEncoder(nje.NumpyConvertEncoder):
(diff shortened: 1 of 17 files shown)


sammy-tri:
It looks like the Mac build has been broken for a while?  Or is there something there I should worry about for this PR?

src/python/director/ikconstraintencoder.py
```diff
@@ -43,25 +38,3 @@ def getPlanPoses(constraints, ikPlanner):
     poses = sorted([c.postureName for c in constraints if hasattr(c, 'postureName')])
     poses = {poseName:list(ikPlanner.jointController.getPose(poseName)) for poseName in poses}
     return poses
-
-
-class IKConstraintEncoder(object):
-    def __init__(self,ikPlanner):
-        self.ikPlanner = ikPlanner
-
-    def publishConstraints(self,constraints,messageName='PLANNER_REQUEST'):
```
mauricefallon: @VladimirIvan : do we use this?

mauricefallon: Aside from the above issue, it seems this is quite minimal renaming.

That said I haven't actually tested it

patmarion:
@sammy-tri fyi have you seen this file? https://github.com/RobotLocomotion/director/blob/master/src/python/tests/testPyDrakeIk.py

mauricefallon:
@wxmerkt  would you mind reviewing or testing this and then accepting it?

src/python/director/ikconstraintencoder.py
wxmerkt: We do not need this as we are handling constraint encoding for EXOTica [here](https://github.com/RobotLocomotion/director/blob/ce215f77ec1aa104e86b56f2b0f0363c219fa691/src/python/director/plannerPublisher.py#L39) using the ``getPlanPoses`` method which remains untouched

wxmerkt:
@sammy-tri you can run individual tests by cd'ing into the directory containing them and using ``directorPython``, e.g. ``directorPython testPyDrakeIk.py``.

PR #302, opened 2016-09-22
adding camera mode to follow/orbit/chase robot

mauricefallon:
previously could only follow ordinary objects e.g. affordances, polydata

demonstration:
https://www.dropbox.com/s/06riyifhrpxmer2/fly-around-1.mp4?dl=0

### DIFF
--- a/src/python/director/cameracontrolpanel.py
+++ b/src/python/director/cameracontrolpanel.py
@@ -5,6 +5,7 @@ from director import visualization as vis
 from director import cameracontrol
 from director import propertyset
 from director import pointpicker
+from director import roboturdf
 
 def addWidgetsToDict(widgets, d):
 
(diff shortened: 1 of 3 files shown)


src/python/director/cameracontrolpanel.py
```diff
@@ -5,6 +5,7 @@
 from director import cameracontrol
 from director import propertyset
 from director import pointpicker
+from director import roboturdf
```
patmarion: unfortunately this file should not import roboturdf, it's new dependency that is too big for the cameracontrolpanel.py.  The camera control panel was designed to be used as a basic component, shouldn't have robot dependencies.  I will make a new pull request to remove this dependency but keeps your new feature.

mauricefallon: that's fine. it should be clear what it is being used to identify that the clicked object is a robot.

You might also want to remove RobotModelFollower as it is replaced by using the common tracker classes

mauricefallon: the video mentioned above reminds me that I split off the maps server from OH so it can be used in Director outside of the OH stack. 

The Cmake still refers to it as "DRC_MAPS" I think

PR #303, opened 2016-09-28
Add dual arm husky

wxmerkt:
Adds package paths and drcargs for dual arm husky


wxmerkt:
@mauricefallon could you please give this a once-over and merge? Thanks

mauricefallon:
@patmarion can you envisage a way to add these modifications from a 'modules' rather that adding this path into the director for each new robot?

I added exactly these two modifications for our basic husky robot a short while back.

PR #306, opened 2016-10-05
Fix camera control panel

patmarion:
remove dependencies from cameracontrol.py and fixes testCameraControl (so travis CI for linux should be green, hopefully)


patmarion:
@mauricefallon next time you update to Director master, please test that the robot camera following had retrained the same behavior.

PR #314, opened 2016-10-11
Fix #313 - glutInit to allow LCMGL text rendering

### LINKED ISSUE #313: lcmgl.text causes director to exit
``` matlab
lcmgl.text(zeros(3,1),'test',0,0)
lcmgl.switchBuffers
```

causes director to close instantaneously with:

```
freeglut  ERROR:  Function <glutBitmapHeight> called without first calling 'glutInit'.
```

wxmerkt:
Pat, thanks for the prompt response and fix - this worked :)

Resolves #313

### DIFF
--- a/src/vtk/DRCFilters/vtkLCMGLProp.cxx
+++ b/src/vtk/DRCFilters/vtkLCMGLProp.cxx
@@ -21,6 +21,7 @@ PURPOSE.  See the above copyright notice for more information.
 
 #include <lcmtypes/bot_lcmgl/data_t.hpp>
 #include <bot_lcmgl_render/lcmgl_decode.h>
+#include <GL/glut.h>
 
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


wxmerkt:
Closing this PR as it can otherwise sometimes lead to ``freeglut (director): illegal glutInit() reinitialization attempt`` - the fix needs to be more elaborate/careful than this

patmarion:
The issue is that lcmgl is mostly self contained without any dependencies, until you try to render text.  For text rendering, the lcmgl-renderer library uses glut, and it expects glut to be initialized.  I would assume it's possible to do this safely, but I've never really looked at how.

patmarion:
If you want, I'd be fine with adding this code, but leave it commented out, so that future users can uncomment and use at their own risk.

was the problem just that the glutInit call was being triggered twice?  You could put a static bool to prevent that from happening.

rdeits:
In case it's helpful, I've used this type of code to deal with a library that needed to be initialized exactly once on startup and freed exactly once on shutdown: 

```c++
namespace {
  struct cdd_global_constants_initializer {
    cdd_global_constants_initializer() {
      dd_set_global_constants();
      std::cout << "Loading cdd global constants" << std::endl;
    }
    ~cdd_global_constants_initializer() {
      std::cout << "Freeing cdd global constants" << std::endl;
      dd_free_global_constants();
    }
  };
  static cdd_global_constants_initializer cdd_init;
}
```

PR #315, opened 2016-10-11
Untangle robotsystem

patmarion:


patmarion:
This is a code refactoring branch.  It shouldn't introduce any behavior changes.

History:

At first, startup.py initialized everything.  Then we started having tests and other programs that wanted to use the core features, but did not want to execute startup.py.  So the core features were moved to robotsystem.py.

This refactors robotsystem.py so that it is a little more manageable.  The new style introduced in this PR isn't perfect, but it's a lot better.  All the code in startup.py could be refactored next to follow this new style.

The code diff is hard to read since almost every line was modified due to indentation.  Here is what robotsystem.py looks like now:

https://github.com/patmarion/director/blob/fb314db9006e3fb76795c94384f82044f3c37186/src/python/director/robotsystem.py

patmarion:
The major difference is that it is much more "plugin" based.  You can import robotsystem and enable just a subset of the components, then initialize it.  It takes the concept of the use<Component> booleans and adds a dependency graph, then initializes components in dependency order.

mauricefallon:
that's a pretty big change. we make heavy use of enabling and disabling components so hopefuly it won't effect us too much.

but happy to take this on if it doesnt change behaviour

patmarion:
yep, I saw that you're now enabling/disabling components via the director_config.json file and setting those in startup.py.  This change is designed to be compatible with that style, I like it.

patmarion:
Here is the big picture:

When someone wants to add a new feature, like a new UI panel, we add it to startup.py.  And that panel might have some required imports, so we add those to the top of startup.py.  And the panel may have some required inputs like the ikPlanner or the footstepsDriver, so we put the initialization code somewhere in startup.py where it can access ikPlanner and footstepsDriver as local variables.

In this new style, you get to explicitly declare: my new component depends on planning and footsteps (and therefore, it does not depend on segmentation, or camera view, etc).  Instead of just plopping it somewhere into startup.py, this new style offers a structured way to introduce new features and declare their dependencies.  It continues to use boolean options like useFoo=True to enable/disable the component (but now we know the dependencies between these).

So I started by refactoring robotsystem.py, since that's a core set of components.  But eventually startup.py will be refactored in this style, too, and put into a module, so other apps can import it.  There are many useful components in startup.py, but other apps can't use them because it's impossible to import just a subset of startup.py.

PR #323, opened 2016-10-14
don't clear pointclouds on load_robot

rdeits:
Currently, a DRAKE_VIEWER_LOAD_ROBOT message deletes the entire drake visualizer folder in the Director object model, which also deletes all point clouds currently being displayed. This change prevents point clouds from being deleted. 

Alternatively, instead of accepting this change, we can decide that the current behavior is actually correct. I'm open to discussion. Should drawing a new robot clear the pointclouds?

### DIFF
--- a/src/python/director/drakevisualizer.py
+++ b/src/python/director/drakevisualizer.py
@@ -362,7 +362,8 @@ class DrakeVisualizer(object):
 
     def removeAllRobots(self):
         for child in self.getRootFolder().children():
-            om.removeFromObjectModel(child)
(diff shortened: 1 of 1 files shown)


patmarion:
I'm fine with having this special case until we decide on a different protocol.  I have no opinions on the protocol right now.  I think everyone agrees the protocol needs to be adjusted to handle more complex scenes, but that's ongoing.

Eventually the pointclouds might also have a frame_id so that they are drawn in the frame of some robot body, maybe those would be in a different folder though.  This seems fine for now.

PR #324, opened 2016-10-17
Refactor ik planner

patmarion:


patmarion:
I'd love to have some folks test this branch with ikPlanner.planningMode = 'exotica'.  As far as I can tell it still publishes the IK request messages exactly how it used to, but I don't have exotica for testing.

Here's the main change:

ikPlanner.plannerPub used to be an instance of PlannerPublisher which was exotica specific.  When ikPlanner.planningMode = 'exotica' then the plannerPub was used.  Now, plannerPub is always used.  But now plannerPub can be a: ExoticaPlannerPublisher, PyDrakePlannerPublisher, or MatlabDrakePlannerPublisher.  When you set ikPlanner.planningMode, the plannerPub automatically switches.

patmarion:
@mauricefallon who would be best to help me test this branch with exotica?  @VladimirIvan?

VladimirIvan:
I'll have a look at it.

patmarion:
I'm going to merge this now, but I won't update oh-distro to use it until we have verified exotica is still working correctly.  If I have broken something, I'm sure it will be easy to fix.

VladimirIvan:
Exotica works fine.
DrakeIK (pydrake) mode fails with the following message though:
```
  File "/home/vivan/dev/oh/software/build/lib/python2.7/dist-packages/director/pydrakeik.py", line 73, in processIK
    endPose, info = self.ikServer.runIk(fields)
AttributeError: 'PyDrakePlannerPublisher' object has no attribute 'ikServer'
```
Do you have a working example of PyDrake? Does this need a newer version of Drake than the one currently in oh-distro-private master?

patmarion:
Great to hear Exotica works after my changes, thanks for testing it!

The PyDrake mode is disabled currently, so the error message you posted is exactly the expected result.  I have a branch where there is still some work remaining before I enable the PyDrake server.  That's here: https://github.com/RobotLocomotion/director/pull/349

patmarion:
By the way, I have added position costs to the planning problem.  In the drake Matlab IKServer, position costs are hard coded, but they should be sent to the ik server along with the planning problem.  So I have added code that makes the default position costs available to the planner publisher, so if you wanted you could add those to your exotica message.

PR #325, opened 2016-10-19
Displaying stereo cloud with limited depth

raluca-scona:
Added capability to display stereo cloud within a set depth. Default range is -1, which does not do any filtering.

### DIFF
--- a/src/app/ddBotImageQueue.cpp
+++ b/src/app/ddBotImageQueue.cpp
@@ -562,7 +562,7 @@ vtkSmartPointer<vtkPolyData> PolyDataFromPointCloud(pcl::PointCloud<pcl::PointXY
 };
 
 //-----------------------------------------------------------------------------
-void ddBotImageQueue::getPointCloudFromImages(const QString& channel, vtkPolyData* polyData, int decimation, int removeSize)
+void ddBotImageQueue::getPointCloudFromImages(const QString& channel, vtkPolyData* polyData, int decimation, int removeSize, float rangeThreshold)
 {
   if (!this->mImagesMessageMap.contains(channel))
   {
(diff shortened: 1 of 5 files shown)


mauricefallon:
Have you done evaluation into how quick this is - compared to not carrying out this filter?

It is possible to apply the threshold in the original conversion library:
https://github.com/openhumanoids/cv_utils/blob/master/multisense_utils/src/multisense_utils/multisense_utils.cpp#L76
This is already doing a loop over the point cloud and accessing the z value so could be much quicker.

raluca-scona:
Hi,

This filter requires around 7ms to run, so it 'feels' real-time. I made a video of what this looks like:

https://www.dropbox.com/s/5xr63kwax1inxpe/demo-depth-filter.mkv?dl=0

I am scrolling through the Max Range parameter in the bottom left corner.

patmarion:
This PR sets up the plumbing to get the max range parameter exposed to the user, very nice.  Agreed that it could probably be faster if the range parameter is passed into the multisense_utils library.  Since it's fast enough for now, I'm happy to accept this PR and we can always speed it up down the road if needed.  Maybe it is fast enough anyway because the parameter reduces the point cloud size.

The ddBotImageQueue class does need some threading to really make it feel snappy.  Right now the image conversion and stereo point cloud conversion both slow down the GUI unnecessarily because it happens on the main thread.  This has been on my todo list forever but has not yet been a priority.

mauricefallon:
I agree. Thanks for the video @raluca-scona . The video is compelling and the computation doesn't seem to be too much.

patmarion:
In order to get this into oh-distro we need to verify director master still works with exotica @VladimirIvan thanks!

PR #328, opened 2016-10-19
adds relative height filtering for lidar

mauricefallon:
for example you can choose to only show lidar 1m below and 2m above the body frame. especially useful for mobile robots where trees and ceilings are a pain as you are usually only interested in the above

demo shows the previous case (with the trees in the way) and then the control dials were added to show only the lidar around the robot's height:
https://www.dropbox.com/s/vghhtrwi7iojuzo/2016-10-lidar_height_control.mp4?dl=0

To be honest I was surprised at the sheer number of changes needed to add this feature


patmarion:
Sorry, haven't had a chance to review this one yet.  Let me know if it's high priority.

mauricefallon:
nope, no rush

patmarion:
Ok, reviewed and ready to merge.  But there are conflicting files, I will fix.

patmarion:
just to clarify, this filters along the world z axis, using the body frame z coordinate in world frame.  That makes sense.  So the sensor can be mounted upside down or vertically, doesn't matter.

mauricefallon:
yes and also so that you dont need to change the bounds when the robot's z varies

PR #331, opened 2016-10-21
automatically provide cmake 3.5 for ubuntu 14 users with older cmake

patmarion:


patmarion:
fyi @wxmerkt we could add a similar thing to the oh-distro build.  This downloads cmake 3.5 and uses it to build lcm and bot_core_lcmtypes.

wxmerkt:
Yes please, that would make things a lot easier!

PR #336, opened 2016-10-21
Adds ability to deactivate instantiation of humanoid task panels with…

wxmerkt:
… one flag

Obviously not the cleanest and best solution (i.e. to move to oh-distro for instance and instantiate for humanoids as via a ``--startup`` flag), but easy change.

@patmarion, would you like to move these modules to oh-distro? Happy to make a pull request and move them including history into e.g. ``software/modules/mitdrc``. Challenge might be that we are currently working with forks so perhaps best to leave them where they are

### DIFF
--- a/src/python/director/startup.py
+++ b/src/python/director/startup.py
@@ -144,6 +144,7 @@ useSpreadsheet = True
 useFootsteps = True
 useHands = True
 usePlanning = True
+useHumanoidDRCDemos = True
 useAtlasDriver = True
 useLCMGL = True
 useOctomap = True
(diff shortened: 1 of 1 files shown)


patmarion:
Let's keep them in director for now.  I have a refactor planned for startup.py.  I haven't began yet, but intend to do it sometime soon.  So please let me know if you guys have pending branches that make mods to startup.py, because we'll end up conflicting.  In the refactor, most of the stuff hanging out in startup.py will get moved into a subdirectory.

patmarion:
when you say "working with forks", you mean oh-distro, right?  Are we all using director master?  Because I intend to make lots of changes to director/master and if you are on a fork I'd be afraid you won't keep up.  I'm hoping you guys are able to pull master and alert me if you find anything broken.  I'm making lots of changes but the intent is not to break anyone's code.

wxmerkt:
We are tracking director on RL/master, I meant the oh forks (RL/valkyrie, OH/oh-distro-private)

PR #338, opened 2016-10-24
Mainwindow app

patmarion:


patmarion:
towards an extensible mainwindow application base.  next steps will be to use this base to implement the drake-visualizer and director applications.

PR #339, opened 2016-10-24
Added logic to fail fast if user wants to use system VTK but version 5 is not available.

### LINKED ISSUE #311: Update Superbuild to Not Default to USE_SYSTEM_VTK=ON
`Director`'s superbuild should check if `vtk5` is installed on the system before it decides to use `USE_SYSTEM_VTK=ON`. If `vtk5` is _not_ installed, `Director` should use `USE_SYSTEM_VTK=OFF`.

This was originally mentioned here: https://github.com/RobotLocomotion/drake/issues/2624#issuecomment-251938584

liangfok:
I tested on this on Ubuntu 16.04 with ROS Kinetic (which has VTK6) and it worked -- the error was printed during the configure stage. I also tested it on Ubuntu 14.04 with ROS Indigo and Director was able to compile using the system VTK.

Resolves #311.

### DIFF
--- a/distro/superbuild/cmake/externals.cmake
+++ b/distro/superbuild/cmake/externals.cmake
@@ -256,6 +256,11 @@ if(NOT USE_SYSTEM_VTK)
   set(vtk_args -DVTK_DIR:PATH=${install_prefix}/lib/vtk-5.10)
   set(vtk_depends vtk)
 else()
+  # Verifies that the system has VTK5.
+  find_package(VTK REQUIRED)
+  if (NOT ${VTK_VERSION_MAJOR} EQUAL 5)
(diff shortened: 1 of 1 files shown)


patmarion:
Thanks!  I guess this might fail on Mac since homebrew vtk is installed to a non-standard location.  Travis-CI will test mac and report back (but sometimes the mac build queue takes a while to start)

If you look in the file just below where you added code, you'll see some special case code to handle homebrew's install location.  I think you could take this homebrew path and add it as a HINTS arg to find_package.  Like:

    find_package(VTK HINTS ${vtk_homebrew_dir})

liangfok:
Oh OK. I'll wait for the current Travis-CI tests to complete to verify it fails before updating this PR with the recommended hint.

patmarion:
Looks good, I think this one will pass the travis tests.

PR #346, opened 2016-10-26
Check whether transform exists in ddBotImageQueue and return if it do…

wxmerkt:
…esnt to avoid segfault

Thanks to this we don't need to define pseudo CAMERA_LEFT frames etc. in robot.cfg files for robots that don't have a multisense

### DIFF
--- a/src/app/ddBotImageQueue.cpp
+++ b/src/app/ddBotImageQueue.cpp
@@ -742,6 +742,10 @@ QList<double> ddBotImageQueue::unprojectPixel(const QString& cameraName, int px,
 //-----------------------------------------------------------------------------
 QList<double> ddBotImageQueue::getCameraFrustumBounds(CameraData* cameraData)
 {
+  // Check whether transform exists and return to avoid segfault
+  if (!cameraData->mCamTrans)
+    return QList<double>();
+
   double width = bot_camtrans_get_image_width(cameraData->mCamTrans);
   double height = bot_camtrans_get_image_width(cameraData->mCamTrans);
   double ray[4][3];


patmarion:
By the way, when we first introduced director_config.json it was only intended to support the different joint names and urdf files of Atlas v3 and Atlas v4.  Then we used it to add support for Valkyrie, then fixed base arms, etc.  But all of this was just used to specify urdf files and kinematic groups for planning.

But now it's clear that we can use it to configure a lot more than robot urdfs.  For example, rather than hardcode the Multisense source and camera names, all of these features could be enabled/disabled from the config.  You guys are already using this idea somewhat to enable/disable code in startup.py, but we could do a lot more.  I think it's on the todo list.  Let me know if you want to chat in person about specific plans.

wxmerkt:
Yes, we are trying to make director more modular and reduce the amount of hacks required to integrate new platforms in terms of bot param and director_config files. Really looking forward to the new startup initialisation :)

Sounds great, happy to have a chat on the roadmap sometime next week :)

mauricefallon:
For our mapping husky, we have a fork of the director with some minor changes.

To give an idea about what's missing for an outdoor navigation / self driving robot, the following changes we made. Ideally director could be configured for a wheeled or quadruped navigation mode without hard coded changes. All other minor modifications live in the extension 'module' (as @patmarion created).

- disabled atlasdriver, irisdriver, handdriver, footstepdriver. And a few derivatives e.g. continuous walking code. 
- generalised the 'footstep goal' to be a 'driving goal' and disabled the footstep planner (which isn't very pretty). Otherwise this provides the same role. (TODO: support moving this on the terrain map.
- there a several irrelevant button on the right side which we havent removed e.g. footstep planning
- enabled moving grid and fixed some bugs in it
- add new map server types - basically different buttons in the maps request panel
- adding robot follow mode. (pushed upstream)
- added a mesh directory to getBuiltinPackagePaths() ... which is necessary for all robots. I think this list should be stored in the json
- added 2 hard-coded lidarDriver instances for two static lidars. This should instead be populated from a list. (bot-viewer does this from the list of lidars in the bot_config)
- more properties become necessary e.g. intensity range for a SICK is way different than hokuyo.
- required drc_lcmtypes to launch due to "import drc" in a variety of unused functions.

In time we will push these mods upstream.

PR #354, opened 2016-10-31
Use RigidBodyTreed typedef 

amcastro-tri:
This PR depends on Drake's PR in RobotLocomotion/drake#3969. That PR must be merged before this one.

This PR simply replaces all the occurrences of `RigidBodyTree` with `RigidBodyTreed`. Per RobotLocomotion/drake#3969, `RigidBodyTreed` simply is a typedef to `RigidBodyTree`.

When RobotLocomotion/drake#3940 the typedef, from Drake's side, will be updated to `RigidBodyTree<double>`.

### DIFF
--- a/src/app/ddDrakeModel.cpp
+++ b/src/app/ddDrakeModel.cpp
@@ -424,7 +424,7 @@ ddMeshVisual::Ptr makeBoxVisual(double x, double y, double z)
   return visualFromPolyData(shallowCopy(cube->GetOutput()));
 }
 
-class URDFRigidBodyTreeVTK : public RigidBodyTree
+class URDFRigidBodyTreeVTK : public RigidBodyTreed
 {
 public:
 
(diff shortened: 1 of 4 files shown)


patmarion:
Please undo your changes to ddDrakeModelOH.cpp, that file shouldn't be modified, thanks!

amcastro-tri:
How so? Otherwise Director won't be able to compile after RobotLocomotion/drake#3940 which then causes  linux-gcc-experimental-ros to fail in Drake's CI. After RobotLocomotion/drake#3940 `RigidBodyTree` will no longer exist. You will either need to use `RigidBodyTree<double>` or its typedef `RigidBodyTreed`

patmarion:
The file ddDrakeModelOH.cpp is only compiled when using an old fork of drake.

amcastro-tri:
I see, I just verified that and reverted to use `RigidBodyTree` (not `RigidBodyTreed`) for that file only.

patmarion:
director is in the unfortunate position of supporting an old drake fork and drake/master.  This is because openhumanoids is stuck on an old drake fork.  All of the incompatibilities are hidden inside ddDrakeModel.  It's important that while you change the api of RigidBodyTree you maintain all the existing funcitonality so that ddDrakeModel can continue to function :)

PR #355, opened 2016-11-01
Revert "Use RigidBodyTreed typedef "

patmarion:
Reverts RobotLocomotion/director#354


wxmerkt:
Just noticed earlier the .h wasn't included in the USE_DRC switch or it didnt work - I was about to look into this; is another attempt at this coming from your end?

patmarion:
There is an open PR in drake to make RigidBodyTree templated, as in, RigidBodyTree<double>.  Once that PR is merged then I will revisit this one.  I'm not thrilled to have to maintain support for OH drake fork and drake/master, but it's the easiest thing for now.  Hopefully that will get us thru 2016 and to a point where relevant APIs in drake/master stabilizes.

PR #357, opened 2016-11-03
Fix default constraint creation for mobile, fixed, humanoid robots

wxmerkt:
- Only create back constraints if the robot has back joints specified in the director config
- Add flag to be able to set ``ikPlanner.robotNoFeet`` from director config via the startup.py component enable/disable

@patmarion would you mind having a look at this whether this is okay/as intended? Using it here to fix teleop planning for a mobile/floating base non-humanoid robot. Tested with val and lwr to verify those still work

### DIFF
--- a/src/python/director/ikplanner.py
+++ b/src/python/director/ikplanner.py
@@ -756,9 +756,7 @@ class IKPlanner(object):
 
 
     def createMovingBodyConstraints(self, startPoseName, lockBase=False, lockBack=False, lockLeftArm=False, lockRightArm=False):
-
-        if (self.fixedBaseArm==False):
-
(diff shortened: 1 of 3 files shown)


patmarion:
Looks fine for now.  This panel needs some TLC to make it not humanoid specific, but still retain all our usefu humanoid features.

Note, there is also a "general purpose end effector teleop panel" which will automatically replace the regular teleop panel if you put config in your director_config.json like this:

   "endEffectorConfig" : {
    "endEffectorLinkNames": ["iiwa_link_ee"],
    "graspOffsetFrame" : [[0,0,0], [0,0,0]],
    "fixedJoints" : ["base_x","base_y","base_z","base_roll","base_pitch","base_yaw"]
  }

PR #358, opened 2016-11-04
Pydrake ik server stable

patmarion:


patmarion:
This is a set of commits that makes progress toward https://github.com/RobotLocomotion/director/pull/349

PR #360, opened 2016-11-05
Add getCameraNames() and only load camera streams if camera defined

wxmerkt:


src/app/ddBotImageQueue.cpp
```diff
@@ -67,26 +67,43 @@ void ddBotImageQueue::init(ddLCMThread* lcmThread, const QString& botConfigFile)
...
+    this->addCameraStream("CAMERA", "CAMERA_LEFT", bot_core::images_t::LEFT);
+  }
+
+  if (std::find(camera_names.begin(), camera_names.end(),
+                "OPENNI_FRAME_LEFT") != camera_names.end()) {
+    this->addCameraStream("OPENNI_FRAME_LEFT");
+    this->addCameraStream("OPENNI_FRAME", "OPENNI_FRAME_LEFT",
+                          bot_core::images_t::LEFT);
+  }
+
```
patmarion: I think all the above code could be removed, since the python layer also calls addCameraStream().  I haven't read closely, but I think this is just old stuff that is ready to be removed.

src/python/director/cameraview.py
```diff
@@ -854,19 +852,30 @@ def cameraFusedCallback(self):
...
-    addCameraView('CAMERACHEST_LEFT', 'Chest left')
-    addCameraView('CAMERACHEST_RIGHT', 'Chest right')
-    addCameraView('CAMERALHAND')
-    addCameraView('CAMERARHAND')
-    addCameraView('KINECT_RGB', 'Kinect RGB')
+    if "modelName" in drcargs.getDirectorConfig():
+        _modelName = drcargs.getDirectorConfig()['modelName']
+        cameraNames = imageManager.queue.getCameraNames()
+
+        if "CAMERA_LEFT" in cameraNames:
```
patmarion: which logic?

src/python/director/cameraview.py
```diff
@@ -717,7 +716,6 @@ def update(self, robotModel):
 
 
 def addCameraView(channel, viewName=None, cameraName=None, imageType=-1):
-
     cameraName = cameraName or channel
     imageManager.queue.addCameraStream(channel, cameraName, imageType)
```
patmarion: addCameraStream() is called here.  The only special case is to handle the "CAMERA" channel, where we need to also call addCameraStream("CAMERA", "CAMERA_LEFT", lcmbotcore.images_t.LEFT)

patmarion:
looks good to me, go ahead and merge it if you want.  i made a few comments about possibly old code that you could remove in this PR if you want, but we can always wait to clean that up later.

PR #387, opened 2016-11-24
changes needed for a non-humanoid, non-OH, non-planning mobile robot - a Husky

mauricefallon:
**This isn't intended to be merged - I'm looking for feedback on should be done with each issue**

The following is a very brief set of changes needed to do the above - which makes the Director fully usable to operate & visualise mobile robots. It is possible with the latest updates from @patmarion 

**Goal**
We don't need/want the following and disabled them in the relevant config.json (which is all fine and expected): planning, DRC demos, IK, AtlasDriver, footsteps

**Issues**
1. I modified addComponent to disable things I don't want, but I do want viewbehaviors 
   - Presumably this can be tailored to a specific robot somehow and not changed like this?
   - I had to make one modification to robotviewbehaviors to remove footstepsDriver

2. A key issue is Iris. It has 3 dependencies and I believe we don't use it:
   - irispy which is used for 1 call each to sample_convex_polytope() and lcon_to_vert()
   - polyhedron - 3 commit repo. Suggestion: insert directly into Director
   - py_drake_utils - only used for one call to rpy2rotmat (which is duplicately implemented in bdi footstep planner)

3. I had to disable this, but would like to keep it: convexhull
I then had to comment out the line in initDefaultOptions calling it 'has no attribute useConvexHullModel'
What is this?

4. Imports of lcmtypes which don't exist:
atlasdriver.py
   - imports atlas lcmtypes
handdriver.py
   - imports irobot and robotiq hand types
Continuous walking demo:
   - imports atlas and ihmc lcmtypes
These could be fixed by using "if available, import"

### DIFF
--- a/src/python/director/atlasdriver.py
+++ b/src/python/director/atlasdriver.py
@@ -18,7 +18,7 @@ import time
 
 import drc as lcmdrc
 import bot_core
-import atlas
+#import atlas
 from pronto.indexed_measurement_t import indexed_measurement_t
 
 
--- a/src/python/director/continuouswalkingdemo.py
+++ b/src/python/director/continuouswalkingdemo.py
@@ -5,7 +5,6 @@ import operator
 import vtkAll as vtk
 import vtkNumpy
 import functools
-import ihmc
 
 from director import lcmUtils
 from director import ioUtils
@@ -26,7 +25,8 @@ import director.tasks.robottasks as rt
 
 import drc as lcmdrc
 import bot_core
-import atlas
+#import atlas
+#import ihmc
 
 from thirdparty import qhull_2d
 from thirdparty import min_bounding_rect
--- a/src/python/director/handdriver.py
+++ b/src/python/director/handdriver.py
@@ -16,8 +16,8 @@ from director.simpletimer import SimpleTimer
 from director.utime import getUtime
 from director import robotstate
 
-import irobothand as lcmirobot
-import robotiqhand as lcmrobotiq
+#import irobothand as lcmirobot
+#import robotiqhand as lcmrobotiq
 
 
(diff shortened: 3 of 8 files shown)


mauricefallon:
FYI: @gtinchev

patmarion:
Thanks for posting, it's good to learn about new use cases.  Maybe we can talk more details on a g+ hangout.

Let me offer you my thoughts on the direction I'd like to go, but note that this direction will require additional development time before it's a workable solution for your use case, and I can't commit to working on this at the moment.

We want to use director in a more minimal environment outside of oh-distro.  So how do we do that?  My proposal: rather than take things away from director, let's start with a minimal app and add features until it is an equivalent replacement to director.  In fact, most of this already exists, it's not yet bundled as a program, like `bin/director`, but you can launch it like this:

    $ alias directorMainWindow='directorPython -m director.mainwindowapp'
    $ directorMainWindow

But now we need to add features to it, similar to the features added by director's startup.py.  With director, you can configure some things, but there's still a lot of hacking/commenting required to strip it down.

So here's an example of how you can start with the minimal `directorMainWindow` and add a robot and perception.  In this example it requires that you execute a custom script, but it could be developed into a standardized feature set that we can configure with a config file.

    $ directorMainWindow --script myApp.py

Where myApp.py contains:

```
from director import robotsystem

factory = robotsystem.RobotSystemFactory()
options = factory.getDisabledOptions()

factory.setDependentOptions(options,
  usePerceptionDrivers=True,
  useSegmentationRobotState=True,
  useSegmentationAffordances=True)

robotSystem = factory.construct(view=view, options=options)
```

You can specify that you want to load the husky robot:

`$ directorMainWindow --script myApp.py --husky`

PR #388, opened 2016-11-28
Generic Drake IK

pvarin:
Adds a generic PlannerPublisher that sends planner requests to an arbitrary IKServer over LCM. Uses LCM types compatible with Drake.


patmarion:
We discussed on slack and decided to move this work into startup scripts until the feature is more developed.

PR #393, opened 2016-12-01
Support adding and removing robots in drakevisualizer

### LINKED ISSUE #389: Drake Visualizer with multiple robots
I've run into an issue that's pretty hard to get around with the way `drake-visualizer` is currently set up. Specifically, I would like to:

* load robot A in drake visualizer once
* at each iteration of my function:
  * draw robot A in a configuration
  * load and draw another robot B (whose geometry changes at every iteration)

I can't currently do that, because loading robot B deletes all robots from the visualizer. I could do:

* at each iteration:
  * load robots A and B
  * draw robot A
  * draw robot B

but it's computationally wasteful, since it loads robot A every time (which might be expensive, if robot A has fancy meshes), and it's inconvenient. 

@patmarion I think we chatted about whether to have the `LOAD_ROBOT` command *only* delete any existing robot with the same robot ID number, rather than all robots. At the time, that seemed like a bad idea, because it leaves additional state in the visualizer. But I'm starting to wonder if it might be the right thing to do. It would certainly fix my immediate problem. 

But perhaps someone else has a better idea, or another use case we need to address? 

How about it? Is anyone else trying to visualize multiple robots simultaneously in the `drake-visualizer`?

rdeits:
Fixes #389 

drakevisualizer now listens to `DRAKE_VIEWER_ADD_ROBOT` ~~and DRAKE_VIEWER_REMOVE_ROBOT~~. `ADD_ROBOT` has the same message type as the existing `LOAD_ROBOT`, but does not clear existing robots (unless an existing robot has the same robot ID number as the robot being added). ~~REMOVE_ROBOT just includes a single robot ID number to delete, or -1 to delete all robots.~~

~~Note: I am abusing the LCM `utime_t` here as the message type for `REMOVE_ROBOT`, in order to avoid creating a branch new LCM type just to send an integer. This feels gross, but I think I'm OK with it~~

### DIFF
--- a/src/python/director/drakevisualizer.py
+++ b/src/python/director/drakevisualizer.py
@@ -304,6 +304,8 @@ class DrakeVisualizer(object):
 
     def _addSubscribers(self):
         self.subscribers.append(lcmUtils.addSubscriber('DRAKE_VIEWER_LOAD_ROBOT', lcmrl.viewer_load_robot_t, self.onViewerLoadRobot))
+        self.subscribers.append(lcmUtils.addSubscriber('DRAKE_VIEWER_ADD_ROBOT', lcmrl.viewer_load_robot_t, self.onViewerAddRobot))
+        self.subscribers.append(lcmUtils.addSubscriber('DRAKE_VIEWER_REMOVE_ROBOT', lcmrl.utime_t, self.onViewerRemoveRobot))
         self.subscribers.append(lcmUtils.addSubscriber('DRAKE_VIEWER_DRAW', lcmrl.viewer_draw_t, self.onViewerDraw))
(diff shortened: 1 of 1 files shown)


patmarion:
looks great.  question- do you actually need the remove single robot feature?  You can always remove all robots by sending a load robot msg with no links.

patmarion:
the reason i ask is because if you dont need the feature then maybe lets hold off on adding it until there is a better message than the utime message.

rdeits:
Yeah, I had considered that as well. Good point on not requiring the weird utime hack. I'll do that.

rdeits:
Ok, done.

rdeits:
@patmarion would you be willing to push a new binary release with these changes?

PR #394, opened 2016-12-01
Add standalone test for drakevisualizer

rdeits:
These commits got lost when #275 was closed. 

This adds a simple standalone python script to demonstrate interacting with the `drake-visualizer` application.


rdeits:
(note that getting CI results back today will be slower than usual due to issues with Travis OSX VMs: https://www.traviscistatus.com/incidents/nq0sf0srvx8s)

patmarion:
great to have this new test, but why all the extra cmake code to run it with 'standalone' python interp?

rdeits:
Uh, I swear there was a reason when I did this back in June, but I can't remember what it was. I'll try getting rid of it to figure out why it's there.

patmarion:
btw, if you run the script with directorPython, you could use that path (sys.argv[0]) to find drake-visualizer.  i see in your cmake code you set the PATH environment variable as another way

rdeits:
Ok, I've removed the standalone cmake logic and used sys.argv[0] to find the drake-visualizer command. Let's see if that works.

rdeits:
@patmarion is investigating why `sys.argv[0]` isn't set in `directorPython`

PR #395, opened 2016-12-04
Updated URDFRigidBodyTreeVTK to Call URDF Parser Directly.

liangfok:
In Drake, I'm working on establishing a layering rule where the URDF and SDF parsers depend on `RigidBodyTree` and not vice versa. For background and motivation, see https://github.com/RobotLocomotion/drake/pull/4317#issuecomment-264149254. This PR modifies Director to conform to this layering design.

### DIFF
--- a/src/app/ddDrakeModel.cpp
+++ b/src/app/ddDrakeModel.cpp
@@ -1,15 +1,10 @@
 #include "ddDrakeModel.h"
 #include "ddSharedPtr.h"
 
-#ifdef DRAKE_OH_FORK
-#include <drake/systems/plants/RigidBodyTree.h>
-#include <drake/systems/plants/shapes/Geometry.h>
-#else
 #include <drake/multibody/joints/floating_base_types.h>
 #include <drake/multibody/parser_urdf.h>
 #include <drake/multibody/rigid_body_tree.h>
 #include <drake/multibody/shapes/geometry.h>
-#endif
 
 #include <vtkPolyData.h>
 #include <vtkAppendPolyData.h>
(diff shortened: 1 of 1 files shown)


patmarion:
Looks good, but could you just remove the ifdef because that guard in not required in his file.  This file is only compiled for drake/master.

liangfok:
OK I will update this PR in about an hour.

amcastro-tri:
@patmarion, I remember adding guards to this same file before (see [line 4 of ddDrakeModel.cpp](https://github.com/liangfok/director/blob/4f1381f0806a5ae0d383389704a41b311552585f/src/app/ddDrakeModel.cpp#L4) ).
I also remember not having to add them to `ddDrakeModelOH.cpp` because that file only got compiled with a very old version of Drake. Did that change?

patmarion:
I think what I said is accurate.  Those guards should be removed too.

liangfok:
I made the requested change. PTAL. Thanks!

liangfok:
Looking in `director/src/app/CMakeLists.txt`, I see the following logic, which indicates that OpenHumanoids uses `ddDrakeModelOH.cpp` instead of `ddDrakeModel.cpp`:

```
  # openhumanoids uses an old drake fork which has a divergent api so we
  # use a fork of ddDrakeModel.cpp to encapsulate the api incompatibilities
  if (USE_DRC)
    set(DRAKE_OH_FORK TRUE)
    list(APPEND srcs ddDrakeModelOH.cpp)
  else()
    set(DRAKE_OH_FORK FALSE)
    list(APPEND srcs ddDrakeModel.cpp)
  endif()
```

Thus, I think it's safe to remove the OH-specific code from `ddDrakeModel.cpp`.

amcastro-tri:
double checking with you @patmarion, since as far as i can remember you asked me to add those guards in the past. Good to know I don't need to worry about them!

patmarion:
Yep.  The guards are needed in the other files, just not ddDrakeModel.cpp.

PR #398, opened 2016-12-07
Refactor pose gui

patmarion:


patmarion:
does anyone still use the `drc-robot-pose-util` bash script?  It either needs to be updated or removed.  I assume most people use the robot pose gui by clicking the button in Director.  But if you need to launch it standalone, the working command line is: `directorPython -m director.robotposegui`

PR #401, opened 2016-12-13
Updated To Use Drake's New Parsers Library and Header File Locations.

liangfok:
These updates are necessary for https://github.com/RobotLocomotion/drake/pull/4482.


patmarion:
I can't merge this as written because it will break in openhumanoids where the parsers library does not exist.  The build logic will have to be expanded to support old/new versions of drake.  I can help advise, but don't have time to work on it myself right now.

liangfok:
Got it. Can you point me to instructions on how to repreduce the build error on the Open Humanoids front?

liangfok:
I will give it a shot today! Thanks.

liangfok:
I looked into the possibility of using `drake-config.cmake` and have a question: Does the OH version of Drake have `drake-config.cmake`? If not, should Director's build system be taught how to search for `drake-config.cmake` and, if the file does not exist, fall back to the existing manual way of specifying the list of Drake libraries to link against?

liangfok:
Does the following website display the Open-Humanoid-pegged versions of Drake and Director?

https://github.com/openhumanoids/oh-distro/tree/master/software

patmarion:
Openhumanoids tracks director/master (but sometimes lags a bit behind).  This is the pegged version of drake: https://github.com/openhumanoids/drake/tree/wxm-fix-oh-drake-for-new-lcm

liangfok:
Thanks, I just cloned and built Open Humanoid's pegged version of Drake and did not find a `drake-config.cmake`. Thus, I believe Director's build system will need to be smart enough to use this file when it exists, but still resort to the current way of finding Drake when this file does not exist.

patmarion:
yes, that's right.  I have a pretty specific plan in mind to do that, eventually.  If you want to take a shot at it, please go for it.  Or replicate the hack I already have for the drakeCommon library.  Note, this only works for additional libraries, but won't work so much for library name changes, etc.  Which is why I wanted to migrate to the new drake-config.cmake.  But I also want to know whether the drake-config.cmake will continue to be support in drake going forward.

liangfok:
Per discussion in Slack / DrakeDevelopers / #bazel, @jwnimmer-tri indicates that `drake-config.cmake` will continue to exist, but downstream projects like Director should eventually link against `libDrake` once https://github.com/RobotLocomotion/drake/issues/4131 is resolved.

@patmarion, given the above, can you provide an outline of your specific plan on how to use `drake-config.cmake`? I can then decide whether I'm able to implement your plan vs. use the aforementioned quick hack.

jwnimmer-tri:
FWIW, I suspect another "quick hack" is the correct answer for this PR (and the concomitant Drake PR), even if a `drake-config.cmake` build system refactor immediately follows.

liangfok:
I went ahead and added the quick hack solution to this PR.

patmarion:
Sounds good.  Please ping me when RobotLocomotion/drake#4482 is ready to be merged this morning and I'll merge this.

jamiesnape:
@patmarion Using `drake-config.cmake` is mostly good to go. The only thing missing is the Eigen include directory is not exported for any of the targets. I am fixing that now.

liangfok:
@patmarion This PR is ready to be merged. https://github.com/RobotLocomotion/drake/pull/4482 passed the feature reviewer and the platform reviewer (@david-german-tri) wants this PR to be merged first.

PR #408, opened 2016-12-19
Create an ubuntu 16.04 build in Docker

### LINKED ISSUE #407: 16.04 binaries
The Travis CI people suggest using docker containers inside Travis as a way to run 16.04 apps. @patmarion does that seem like a reasonable way to auto-generate the 16.04 binaries?

rdeits:
~~This is a work in progress (not ready for merge yet).~~

Travis doesn't support 16.04 yet, but we need a build to generate the prebuilt binaries. We can run one inside a Docker container on Travis. 

Fixes #407


rdeits:
@patmarion I think I've got the build and tests working inside the 16.04 docker image. I don't actually know how the package upload process works, but I think it should be pretty easy from here. Can you look at it with me tomorrow?

patmarion:
Looks like everything is working, very nice!

The package upload will not succeed from a PR from your fork because you cannot access the encrypted variable that stores ssh keys.

The last thing to do before merging is to put in a guard so that packaging is only run for branch master and not for each PR.

patmarion:
See the very end of this script, to see how the commands to make package is skipped for pull requests:

https://github.com/RobotLocomotion/director/blob/master/distro/travis/after_success.sh#L65

rdeits:
Okay, I think all of the relevant variables are now getting passed in as Docker build args. Note that this essentially bakes your Travis secrets into the resulting Docker image. That's fine right now, since the image itself is destroyed when Travis finishes. But if we ever decide to publish a base image with the dependencies included, we need to make sure that the base Dockerfile does not have any of these ENV or ARG lines.

rdeits:
Hooray! It passes! Should be good to go.

PR #419, opened 2017-01-05
Shiny new drake visualizer

### LINKED ISSUE #415: New drake visualizer interface
Continuing the discussion from https://github.com/rdeits/DrakeVisualizer.jl/issues/21#issuecomment-268728737

I'd like to propose a complete overhaul of the drake visualizer interface. The current drake-visualizer is great, but it has several issues that can't be avoided with its current architecture:

* clients must always send a LOAD message before sending a DRAW message. If the visualizer is ever restarted, the clients have no way of knowing, and will just keep sending useless DRAW messages. 
* the only organizational structure allowed is an integer robot_id_number and a string link name
* the only way to add new geometries or add new features to existing geometries is to pack floats into the float_data of the `viewer_geometry_data_t` lcmtype, which is extremely clunky
* drake-visualizer currently uses entirely different channels and code paths to draw
  * link geometries
  * lidar points
  * pointclouds
* that means that, for example, several lidars cannot be rigidly connected, and a pointcloud cannot be attached to a robot link

To solve all of these issues, I'm proposing several major changes:

* Create a response protocol for the visualizer, which at least indicates whether a given action (like DRAW or LOAD) was successful. 
  * More specifically, when a client tries to draw a link at a given pose, the visualizer may respond with a "no such link" message and the name of that link, so that the client can send an appropriate LOAD message
* Stop using LCM types for geometry data. The fact that new fields cannot be added to LCM types without breaking compatibility is a major annoyance. Instead, we can just pack a set of bytes that encode, e.g. a JSON string into an LCM message. This makes it trivial to add new options in the future, without breaking compatibility
* Use this increased type flexibility to make pointclouds and lidars just another geometry type. This makes it easy to add pointclouds which rigidly move together (by attaching them to the same link)
* Replace the "robot_id_num" and link name concepts with a general "path" concept. A path is a list of strings, and a link with path ["robot1", "group2", "link3"] is placed in the folder `robot1/group2/link3` in the visualizer. This reproduces all of the capability of the robot_id_num concept and also allows lots of other organizational schemes

I've done all of the above on my branch here: https://github.com/rdeits/director/tree/json-interface

Pointclouds, meshes, LIDARs, and geometric primitives are all working. The basic client-server protocol is working, so now the visualizer can be killed and restarted while a client is drawing, and everything will "just work" (i.e. the model will be re-loaded when necessary). It also makes most of the drake visualizer code quite a bit simpler, since we don't have separate code for pointclouds and lidars. 

It's possible to use the new interface without the client being smart: if you just blindly send LOAD and then DRAW (the way the clients do now), everything will work. You just won't have the benefit of being able to stop and start the viewer at will. 

But before I PR this, I want to make sure this is the right thing to do. I don't imagine we'll want to delete the current drake-visualizer. Instead, I think we should keep it (but deprecate it) and bring in the new interface as `drake-visualizer2` or similar. 

Also, I haven't noticed any performance issues from the switch from LCM types to JSON dicts. But if there is one, we can trivially swap out JSON for MsgPack with almost no code change (and we can support clients using both formats with no trouble).

rdeits:
@patmarion this will probably need some work before it's ready to merge, but I wanted to get started on the PR now. 

This implements the interface described in #415 (closes #415)

I've split the old `drakevisualizer.py` into `drakevisualizer.py` and `drakevisualizerapp.py`. I've also created the new interface in `drakevisualizer2.py`. Additionally, I've created `drakevisualizerapp1.py` and `drakevisualizerapp2.py` which import the relevant modules to construct the v1 and v2 apps. And I've added a command line flag to `drakeVisualizerApp.cpp`

@patmarion I have no attachment to any of those naming choices. They're just the first thing I tried that managed to cleanly separate all the old and new functionality. Feel free to rearrange and rename them as you see fit. Everything I've added is contained in `drakevisualizer2.py` and the new `testDrakeVisualizer2Interface.py`; all the other changes are just rearrangements of existing code.

### DIFF
--- a/distro/superbuild/cmake/externals.cmake
+++ b/distro/superbuild/cmake/externals.cmake
@@ -146,7 +146,7 @@ if (USE_LCM)
 
   ExternalProject_Add(bot_core_lcmtypes
     GIT_REPOSITORY https://github.com/rdeits/bot_core_lcmtypes
-    GIT_TAG 9446ce422b679e88968718f63715c681cf621b70
+    GIT_TAG dd1236b5bf5f2662e200d78297393333fe97ae64
     ${cmake3_args}
     CMAKE_CACHE_ARGS
       ${default_cmake_args}
--- a/distro/superbuild/cmake/externals.cmake
+++ b/distro/superbuild/cmake/externals.cmake
@@ -145,8 +145,8 @@ endif()
 if (USE_LCM)
 
   ExternalProject_Add(bot_core_lcmtypes
-    GIT_REPOSITORY https://github.com/openhumanoids/bot_core_lcmtypes
-    GIT_TAG 9967654
+    GIT_REPOSITORY https://github.com/rdeits/bot_core_lcmtypes
+    GIT_TAG 9446ce422b679e88968718f63715c681cf621b70
     ${cmake3_args}
     CMAKE_CACHE_ARGS
       ${default_cmake_args}
--- a/src/app/drakeVisualizerApp.cpp
+++ b/src/app/drakeVisualizerApp.cpp
@@ -8,7 +8,13 @@ int main(int argc, char **argv)
   ddPythonManager* pythonManager = new ddPythonManager;
   pythonManager->setSysArgv(QApplication::instance()->arguments());
   PythonQt::self()->addVariable(PythonQt::self()->importModule("sys"), "executable", QCoreApplication::applicationFilePath());
-  pythonManager->executeString("import director.drakevisualizer; director.drakevisualizer.main(globals())");
+
+  if (argc > 1 && strcmp(argv[1], "-v2") == 0) {
+    pythonManager->executeString("import director.drakevisualizerapp2; director.drakevisualizerapp2.main(globals())");
+  } else {
+    pythonManager->executeString("import director.drakevisualizerapp1; director.drakevisualizerapp1.main(globals())");
+  }
+
 
   // delete pythonManager;
   // Allow a leak to avoid a segfault in the PythonQt cleanup destructor.
--- a/newviewer.py
+++ b/newviewer.py
@@ -4,7 +4,7 @@ import time
 import math
 import sys
 import numpy as np
-sys.path.append("build/install/lib/python2.7/site-packages")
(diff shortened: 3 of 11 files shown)


distro/superbuild/cmake/externals.cmake
```diff
@@ -145,8 +145,8 @@ endif()
 if (USE_LCM)
 
   ExternalProject_Add(bot_core_lcmtypes
-    GIT_REPOSITORY https://github.com/openhumanoids/bot_core_lcmtypes
-    GIT_TAG 9967654
+    GIT_REPOSITORY https://github.com/rdeits/bot_core_lcmtypes
+    GIT_TAG dd1236b5bf5f2662e200d78297393333fe97ae64
```
rdeits: we should PR the new LCM type into mainline bot_core_lcmtypes before merging this

src/app/drakeVisualizerApp.cpp
```diff
@@ -8,7 +8,13 @@ int main(int argc, char **argv)
...
   pythonManager->setSysArgv(QApplication::instance()->arguments());
   PythonQt::self()->addVariable(PythonQt::self()->importModule("sys"), "executable", QCoreApplication::applicationFilePath());
-  pythonManager->executeString("import director.drakevisualizer; director.drakevisualizer.main(globals())");
+
+  if (argc > 1 && strcmp(argv[1], "-v2") == 0) {
+    pythonManager->executeString("import director.drakevisualizerapp2; director.drakevisualizerapp2.main(globals())");
+  } else {
+    pythonManager->executeString("import director.drakevisualizerapp1; director.drakevisualizerapp1.main(globals())");
+  }
+
```
rdeits: I added a runtime switch to the `drake-visualizer` app. But we could also just produce two separate binaries. I have no particular feelings about either approach.

patmarion:
Are you still adding to this branch?  If you can take a break from it and let me do a pass over it, I may have time to do a pass over the branch.  I might remove the numbers from the module names and do some rename/reorg.

rdeits:
I don't have any immediate changes planned, so it's all yours. And yes, removing the numbers and renaming sounds great.

PR #420, opened 2017-01-06
Shiny new drake visualizer

### LINKED ISSUE #415: New drake visualizer interface

rdeits:
Copied from #419 so @patmarion can also push to the branch. 

@patmarion this will probably need some work before it's ready to merge, but I wanted to get started on the PR now. 

This implements the interface described in #415 (closes #415)

I've split the old `drakevisualizer.py` into `drakevisualizer.py` and `drakevisualizerapp.py`. I've also created the new interface in `drakevisualizer2.py`. Additionally, I've created `drakevisualizerapp1.py` and `drakevisualizerapp2.py` which import the relevant modules to construct the v1 and v2 apps. And I've added a command line flag to `drakeVisualizerApp.cpp`

@patmarion I have no attachment to any of those naming choices. They're just the first thing I tried that managed to cleanly separate all the old and new functionality. Feel free to rearrange and rename them as you see fit. Everything I've added is contained in `drakevisualizer2.py` and the new `testDrakeVisualizer2Interface.py`; all the other changes are just rearrangements of existing code.


rdeits:
@patmarion I added one commit to support the optional "transform" attribute for geometries: 972a6ee

rdeits:
@patmarion FYI I pushed one more commit to remove some duplication from the status messages

rdeits:
@patmarion so far, performance is OK with this change, but not great. The biggest slowdown turns out not to be JSON-related, but just due to searching through the objectmodel. When a client sends something like `draw path ["foo", "bar", "baz"] at transform T`, director has to search through the objectmodel for folder foo, then for its child bar, then for its child baz. This turns out to be surprisingly slow (about 300 microseconds per robot link), especially for complicated robots. Is there any way for me to access folders in the objectmodel more quickly than just doing a bunch of `findChild` calls? 

If I have to cache these lookups I will, but I'd really rather avoid introducing all of the obvious problems that caching always creates.

patmarion:
The object model cannot provide speedy name lookups because:

  - item names do not have to be unique
  - item names may be changed at any time

However, the rules of the drawing protocol say that names are unique (since it is path based) so I would recommend that the drawing class caches the items in a python dict by pathname, then let's discuss where that could be buggy.

rdeits:
Ok, sounds good. The particular issue I ran into when I tried that was that it's easy to keep a cache inside the `DrakeVisualizer` class and remove elements when the client sends a `delete` message. But we also have to ensure that the cache is cleared when the user does something like manually delete a folder from the objectmodel in director.

patmarion:
There is a callback you can use whenever an item is removed from the object model.  It is fired exactly once as the item is removed.  Triggered for each item whether an item is removed programmatically or by the user pressing 'delete' on an item or folder.

```

def onItemRemoved(self, objModel, item):
    # remove item from cache

def addItem(path, item):
  self.cache[path] = item
  item.connectRemovedFromObjectModel(self.onItemRemoved)

```

rdeits:
Ok, thanks, that's perfect.

rdeits:
@patmarion I'm making the minor changes to the way format versions are specified that we talked about today. Also, we should remember to deal with the new LCM type that I added to bot core (this PR currently points Director's build to my fork of bot core types)

rdeits:
Also, I'm using the `viewer2` name in a few places in code to refer to this version of the visualizer. If we want to give it a better name, now would be a good time.

PR #422, opened 2017-01-09
Updated ddDrakeModel.cpp to use RigidBodyTree's new API where the model instance ID set needs to be explicitly specified.

liangfok:
This needs to go in alongside https://github.com/RobotLocomotion/drake/pull/4706.


liangfok:
I just had a F2F conversation with @jwnimmer-tri and his recommendation is to not change RigidBodyTree's API for the time being. I will update https://github.com/RobotLocomotion/drake/pull/4706 do not change the API.

PR #423, opened 2017-01-12
Add home nominal and clip plans to avoid numerical precision error when converting float to double

VladimirIvan:
* Added an option to compute home nominal plans without moving the arms.
  - Such function used to exist in the past.
* Added clipping of ik poses and ik trajecotries to joint limits that take into account 32bit floating point rounding errors.
  - The plan gets sent via LCM as float (32bit) which reduces the accuracy. This causes problems when the joints limits are checked in the controller against a joint limit at 64bit double accuracy. The difference is small (1e-6) but it causes the joint limit checks to fail. This fix reduces the joint limits by 1e-6 and clips the plans to this joint limit. This fix could exist in the translator or the controller as well. I put it into the ik planner so that the plan can be visualized.

### DIFF
--- a/src/python/director/ikplanner.py
+++ b/src/python/director/ikplanner.py
@@ -73,6 +73,7 @@ class ConstraintSet(object):
         self.endPose, self.info = self.ikPlanner.plannerPub.processIK(self.constraints, ikParameters, positionCosts, nominalPoseName=nominalPoseName, seedPoseName=seedPoseName)
 
         self.endPose = self.ikPlanner.clipState(self.endPose,self.jointLimitsLower,self.jointLimitsUpper) 
+
         self.ikPlanner.addPose(self.endPose, 'q_end')
         print 'info:', self.info
         return self.endPose, self.info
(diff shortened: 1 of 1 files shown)


src/python/director/ikplanner.py
```diff
@@ -1478,6 +1504,7 @@ def runIkTraj(self, constraints, poseStart, poseEnd, nominalPoseName='q_nom', ik
         ikParameters = self.mergeWithDefaultIkParameters(ikParameters)
 
         self.lastManipPlan, info = self.plannerPub.processTraj(constraints, ikParameters, positionCosts, nominalPoseName=nominalPoseName, seedPoseName=poseStart, endPoseName=poseEnd)
+        self.lastManipPlan = self.clipPlan(self.lastManipPlan)
```
patmarion: i think that this plan object isn't actually the plan object that it published when a plan is committed.  There is a separate class which receives plan messages (CANDIDATE_MANIP_PLAN) and assigns those plans to the plan playback panel.  When you click "Execute" then the plan is published on COMMITTED_MANIP_PLAN.  For example, think of LCM log playback, plans come from LCM, not from the ikplanner.  This is a confusing aspect and sometimes, we programmatically commit plans from the ikplanner, so this is something we should probably fix.

patmarion:
hey if it's not too much trouble you might want to open a separate PR for the clipping than for the home plan.  I think we can accept the change for the home plan right away, but for clipping i think we should discuss it a bit.

questions:

- perhaps clipping should not be enabled by default, it is a slight layer of complexity/surprise that might not be desired by all users

- i can see arguments for handling the clipping at a different level, but also see good reasons to handle it here in the application layer.  just want to make sure we are handling it in the application layer because that is best, and not because that is easiest/quickest.

patmarion:
do you mind if i nitpick on your python style? I'd like to have that cleaned up before merging.  Particularly, in some places you are missing spaces after commas, spaces around equal signs, and used 2 instead of 4 spaces for indents.

VladimirIvan:
Splitting this into separate pull requests

PR #425, opened 2017-01-13
Clip plans to joint limits before displaying them to avoid float rounding issues

VladimirIvan:
This option works and shows the augmented plan before it gets executed.
It does not account for joint limits being temporarily adjusted using the joint limit warning button in the status bar.

It's turned off by default.

This fix could live in the translator as well. Please discuss.


patmarion:
Rather than testing the conditional `self.clipFloat32SafeJointLimits` inside the clipState and clipPlan functions, I think it would be better for the caller to test and conditionally call the functions.  The functions themselves should always have predictable behavior.

Also, I'm still not convinced this is clipping the visualized plan.  I think the plan playback panel gets plans from a different lcm subscriber and not the ikplanner.  Am I wrong?

patmarion:
Note, the plan message that is clipped is also committed here:

https://github.com/oh-dev/oh-distro-private/blob/a1ce7c9799204fc509d9ef7f055932af90da529d/software/modules/ohdirector/tableboxdemo.py

However, the visualized plan, and the plan that is committed if you click "Execute" in the plan playback panel, is *a different* plan message.  The plan message returned from ikPlanner is received by one lcm subscriber, but there is *a different* subscriber that receives plans and passes them to the plan playback panel.  We could fix this, so that there is exactly *one* subscriber that ever receives messages on the CANDIDATE_MANIP_PLAN channel, but currently this is not the case.

src/python/director/ikplanner.py
```diff
@@ -1474,6 +1482,23 @@ def mergeWithDefaultIkParameters(self, ikParameters):
...
         return newIkParameters
 
+    def clipState(self, q, l, u):
+        state = np.array(q)
+        ret = np.clip(state, l, u)
+        if np.max(np.abs(state-ret)) > 1e-6:
+            print 'State is outside of joint limits! clipped to limits.'
+            return ret.tolist()
+        else:
+            return ret.tolist()
```
patmarion: one nitpick, would you mind updating this function for readability:

 - rename function and args to "clipJointPositions(self, jointPositions, lowerBound, upperBound):"
 - remove the else condition and un-indent the return statement.
 - don't create a new variable name, the first line could just be `jointPositions = np.array(jointPositions)`

VladimirIvan: Done

PR #426, opened 2017-01-13
Resolve #313 - initialisation of GLUT if it hasn't been previously done

### LINKED ISSUE #313: lcmgl.text causes director to exit

wxmerkt:
Second attempt at #313 after #314 - this one checks if GLUT has previously been initialised and only initialises when it hasn't been.

### DIFF
--- a/src/vtk/DRCFilters/CMakeLists.txt
+++ b/src/vtk/DRCFilters/CMakeLists.txt
@@ -130,6 +130,11 @@ if (USE_DRC_PLANE_SEG)
 endif()
 
 if(USE_LCMGL)
+  # Check whether GLUT is available to enable/disable some LCMGL-features
+  find_package(GLUT)
+  if(GLUT_FOUND)
+    add_definitions(-DGLUT_AVAILABLE)
+  endif()
 
   find_package(LibBot REQUIRED)
   include_directories(${LIBBOT_INCLUDE_DIRS})
--- a/src/vtk/DRCFilters/vtkLCMGLProp.cxx
+++ b/src/vtk/DRCFilters/vtkLCMGLProp.cxx
@@ -21,7 +21,11 @@ PURPOSE.  See the above copyright notice for more information.
 
 #include <lcmtypes/bot_lcmgl/data_t.hpp>
 #include <bot_lcmgl_render/lcmgl_decode.h>
+
+// lcmgl.text requires GLUT at runtime, cf. #313
+#ifdef GLUT_AVAILABLE
 #include <GL/glut.h>
+#endif
 
 //----------------------------------------------------------------------------
 vtkStandardNewMacro(vtkLCMGLProp);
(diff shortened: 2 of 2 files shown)


patmarion:
This adds a new dependency on GLUT, but only when LCMGL is enabled.  Does libbot's lcmgl already depend on GLUT, that is, is it safe to assume that LCMGL=ON implied GLUT is available?

wxmerkt:
Good question - bot-vis depends on GLUT, ``lcmgl_decode`` depends on bot-vis. So I'd say there already is a GLUT dependency when using LCMGL, even though it only gets triggered with some of the calls that require it

patmarion:
i'll merge this now if you disable it by default in the c++ file using \#if preprocessor macros.  I think it will take some extra cmake logic to get the GLUT dependency to work.  here is a travis build error, where lcmgl is on, yet director does not have GLUT:

https://s3.amazonaws.com/archive.travis-ci.org/jobs/191665641/log.txt

wxmerkt:
Awesome, thanks, I've wrapped it based on whether CMake can find GLUT or not, is this along the lines you were thinking? Let's see what Travis says :)

patmarion:
cool, yeah i can merge this if travis likes it :)

patmarion:
sorry, forgot to come back to this one!

jamiesnape:
GLUT on OS X is not at `<GL/glut.h>`. You need `<GLUT/glut.h>` and `<OpenGL/OpenGL.h>`.

patmarion:
@wxmerkt sorry I have to revert this.  Turns it out wasn't tested properly on travis-ci's mac image, bummer.   I thought that GLUT was installed on the travis-ci image and that this was tested, but I guess not.

I'm happy to add the feature back, but rather than it being conditional on FindPackage(GLUT), I'd like to add a cmake option like option(USE_GLUT_GLTEXT OFF) which defaults to OFF.  The problem is some people might have GLUT installed but not actually want Director to use it.

wxmerkt:
Can we just have a preprocessor directive to check if it's Apple to switch between the two?

patmarion:
Yes, that would be the correct way to resolve the issue about the include statement.

I would still request an explicit cmake option to enable/disable the functionality.  Then we can debate over whether it should be enabled by default or not.  I'd prefer to have the option, instead of the current implementation which is always enabled if GLUT is present.

patmarion:
see, for example, https://github.com/RobotLocomotion/director/blob/master/src/common/ddOpenGL.h

jamiesnape:
Using a header generated using `configure_file` might be preferable to using `add_definition` too.

wxmerkt:
Okay, thanks for the input. It's not a high priority for me so happy to leave it as "won't fix/support lcmgl-text" (it mitigates a runtime segfault).

PR #429, opened 2017-01-17
Added try except statements for atlas, iris, drake's util and ihmc modules

gtinchev:


patmarion:
I understand that this is a quick fix to use Director in a configuration with some module dependencies (mostly lcm message types) removed.  But I don't really want to implement it this way, by just hacking out import statements with try/except.  I've already done quite a lot to support optional components in robotsystem.py.  But, this work is not complete, and not integrated into Director's startup.py.  I'm interested to see it finished but unfortunately it hasn't been a priority.

Here is a quick example I just made that shows how to prevent modules like handdriver.py and atlasdriver.py from being imported.  This is just a quick example and not something I can merge:

https://github.com/RobotLocomotion/director/compare/master...patmarion:disable-components-quick

As shown in the example, rather than put try/except into modules, I'd prefer to prevent the modules from being imported in the first place.  However, the better implementation of the above example would be to make startup.py use the `ComponentFactory` style that is used in robotsystem.py.  I still hope to get around to that eventually.

PR #437, opened 2017-01-23
Support drawing triads in the new TreeViewer

rdeits:
cc @dehann

### DIFF
--- a/src/python/director/treeviewer.py
+++ b/src/python/director/treeviewer.py
@@ -484,12 +484,11 @@ class TreeViewer(object):
                 om.removeFromObjectModel(existing_item)
             else:
                 item.setProperty("Point Size", 2)
+                availableColorModes = set(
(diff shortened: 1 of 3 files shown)


src/python/director/treeviewer.py
```diff
@@ -476,7 +484,7 @@ def addGeometry(self, path, geomItems):
                 om.removeFromObjectModel(existing_item)
             else:
                 item.setProperty("Point Size", 2)
-                for colorBy in ["rgb", "intensity"]:
+                for colorBy in ["rgb", "intensity", "Axes"]:
```
patmarion: btw, item.getPropertyAttribute('Color By', 'enumNames') will return the list of available color by names, in case you want to use that.

rdeits: fixed, thanks

dehann:
@rdeits , got it thanks!

PR #438, opened 2017-01-23
launch viewer window in tree viewer test

rdeits:
This code is copied over from the drake interface test. It could also be refactored out, but I'm somewhat more inclined towards code reuse within tests than I would otherwise be.


patmarion:
nice!

PR #439, opened 2017-01-25
Add panels

patmarion:


patmarion:
CI failed on mac due to issue with homebrew install vtk5.

PR #443, opened 2017-01-27
Director updates

patmarion:


patmarion:
Tested with atlas v5 on DRC log.

PR #448, opened 2017-02-07
first draft of python viewer client interface

rdeits:
~~patmarion this is just the beginning (and not ready to merge yet), but I wanted to open the PR so you can see what I'm doing here~~

(see the end of the file for some basic usage)

Todo: 
- [x] demonstrate lists of objects
- [x] demo multiple objects with distinct colors that show up as the same objmodel item
- [x] inherit from object where appropriate
- [x] add a test case that uses this
- [ ] add response handler
- [x] handle path as string

### DIFF
--- a/src/python/director/viewerclient.py
+++ b/src/python/director/viewerclient.py
@@ -93,6 +93,30 @@ class CommandQueue:
         return not (self.draw or self.load or self.delete)
 
 
+class Visualizer:
+    __slots__ = ["core", "path"]
+    def __init__(self, core=None, path=None, lcm=LCM()):
+        if core is None:
+            core = CoreVisualizer(lcm)
+        if path is None:
+            path = tuple()
+        self.core = core
+        self.path = path
+
+    def load(self, geomdata):
+        self.core.load(self.path, geomdata)
+        return self
+
+    def draw(self, tform):
+        self.core.draw(self.path, tform)
+
+    def delete(self):
+        self.core.delete(self.path)
+
+    def __getitem__(self, path):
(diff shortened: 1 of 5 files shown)


rdeits:
Ok, I think this is at a state where it would actually be useful to have. @patmarion can you please take a look?

src/python/director/viewerclient.py
```diff
@@ -0,0 +1,240 @@
...
+        return self
+
+    def draw(self, tform):
+        self.core.draw(self.path, tform)
+
+    def delete(self):
+        self.core.delete(self.path)
+
+    def __getitem__(self, path):
+        return Visualizer(self.core, self.path + (path,))
```
patmarion: should also pass lcm instance? `lcm=self.core.lcm` ?

rdeits: Yeah, that's probably better (it currently has no effect, since that `lcm` arg is only used if `core` is not passed, but "explicit is better than implicit")

patmarion: oh maybe really what is better is for the constructor  default arg to be lcm=None and only ever call LCM() when needed.

rdeits: it actually only ever happens once (when the method is defined, not when it's called), but I guess we could get that down to zero times by using None as the default. Ok.

rdeits:
okay, multi-geometry paths are now working pretty nicely. @patmarion I changed the Geometry class to use a `vtkAppendPolyData` to smoosh all the data together instead of creating separate objectmodel items.

patmarion:
merged initial set, even though there's probably more changes to add.

PR #453, opened 2017-02-09
rename load -> setgeometry and draw -> settransform and fix polyDataItem properties

### LINKED ISSUE #451: fix incorrect color mode on re-loaded geometry in viewerclient

### LINKED ISSUE #452: rename `load()` and `draw()` to `setgeometry()` and `settransform()` in the tree viewer interface

rdeits:
Deprecates the old `load()` and `draw()` commands, so they'll still work for now but will issue warnings. 

Fixes #452 

Also creates fewer unnecessary polyDataItems, which means that copying properties from one polyDataItem to another is no longer necessary, which fixes #451

### DIFF
--- a/src/python/director/treeviewer.py
+++ b/src/python/director/treeviewer.py
@@ -1,13 +1,14 @@
+import json
 import math
 import os
-import json
 import time
+import warnings
 import numpy as np
 from collections import namedtuple
-from director import lcmUtils
 
-import director.objectmodel as om
-import director.applogic as app
+from director import objectmodel as om
+from director import applogic as app
+from director import lcmUtils
 from director import transformUtils
 from director.debugVis import DebugData
 from director import ioUtils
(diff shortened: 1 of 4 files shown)


rdeits:
sorry, this is hard to disentangle from my changes to address #451

rdeits:
Ok, both fixes are just coming in together, sorry.

src/python/director/treeviewer.py
```diff
@@ -326,26 +320,38 @@ def addColorChannels(polyData, channels):
             colorArray = np.asarray(channels["rgb"]) * 255
             vnp.addNumpyToVtk(polyData, colorArray.astype(np.uint8), "rgb")
 
-    def __init__(self, geomDatas, polyData):
-        self.polyDataItem = vis.PolyDataItem("geometry", polyData, view=None)
-        self.polyDataItem._updateColorByProperty()
+    def __init__(self, geomDatas):
```
rdeits: previously, constructing a Geometry would also construct a new polyDataItem. That's not necessary when there's an existing item at the target path, so now the Geometry just owns a polyData, not a polyDataItem.

PR #463, opened 2017-02-15
Add a unique client_id to the channel name in the treeviewer protocol

rdeits:
cc @patmarion 

Clients should send requests on `DIRECTOR_TREE_VIEWER_REQUEST_<foo>` and listen on `DIRECTOR_TREE_VIEWER_RESPONSE_<foo>`. The old behavior will still work, but raises a deprecation warning. 

This also updates `viewerclient.py` to follow the new protocol (using a unique temporary filename as its client ID). `testTreeViewerInterface.py`, however, is still testing the deprecated protocol (for now).

### DIFF
--- a/src/python/director/viewerclient.py
+++ b/src/python/director/viewerclient.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 import time
 import json
+import os
 import tempfile
 import threading
 from collections import defaultdict, namedtuple, Iterable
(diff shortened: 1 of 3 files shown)


src/python/director/viewerclient.py
```diff
@@ -205,13 +206,20 @@ def __init__(self, lcm=None):
         if lcm is None:
             lcm = LCM()
         self.lcm = lcm
+        self.client_id_file = tempfile.NamedTemporaryFile()
```
patmarion: i don't love this.  what about `os.getpid()` plus a global counter variable, or id(self).  (trying to think of short strings, uuid is probably the best long string)

rdeits: Ok, PID + counter is fine with me.

rdeits: fixed

patmarion:
lgtm, ready to merge?

rdeits:
yup

PR #466, opened 2017-02-16
use git describe string for package naming

patmarion:


patmarion:
@rdeits this will change the naming to something that looks like:

director-0.1.0-2-g8cce081-ubuntu-16.04.tar.gz

getting a bit long, but so it goes.  Do you have any preference between 0.1.0 and v0.1.0?  That is determined by the tag name that I pushed to origin.  Seems there are plenty of examples of both forms.

Also note that the git describe string prefixes the sha1 with a "g" as in  "g<sha1>".

rdeits:
Ok, sounds good. I have no preferences on naming either way; I'm happy with anything. 

Do you mind keeping the `director-0.1.0-ef7d121` binaries around with their current name? I've already submitted a release of DrakeVisualizer.jl pointing to that version.

patmarion:
yep i'll leave those binaries. let me know in the future when they can be removed.

rdeits:
Actually, there's still time to change DrakeVisualizer.jl before the new version is released. If you can get the renamed binaries up for that version now, then we can delete them today.

PR #445, opened 2017-02-01
add utils to ddDrakeModel

manuelli:
- Add some utilities to ddDrakeModel.
- Add utilities to python side of director.

@patmarion could you review?

PR #469, opened 2017-02-17
Revert "add utils to ddDrakeModel"

patmarion:
Reverts RobotLocomotion/director#445


patmarion:
sorry @manuelli, found another bug and i need to revert this until we can get it fixed.

manuelli:
@patmarion is the bug in #445? If you open an issue I can tackle it this week.

PR #486, opened 2017-03-05
Lm cpf director merge

patmarion:


src/app/ddDrakeModel.h
```diff
@@ -78,6 +84,10 @@ class DD_APP_EXPORT ddDrakeModel : public QObject
   static void addPackageSearchPath(const QString& searchPath);
   static QString findPackageDirectory(const QString& packageName);
 
+  #ifdef DRAKE_OH_FORK
+  void doKinematics(const QVector<double>& q, const QVector<double>& v, bool compute_gradients = false, bool compute_JdotV = true);
```
patmarion: are you using this? can it be removed?

manuelli: I'm not using this currently. I think it would be nice to be able to call doKinematics with positions and velocities. Currently you can only do it with positions, through the setJointPositions method, this implicitly sets all velocities to zero. I'm not sure worried about supporting this use case however since really people should directly use the drake python api instead of going through director.

patmarion: yeah i would prefer to limit the scope of the api of ddDrakeModel now that the drake api is directly available in the python bindings

manuelli:
Cleanup up some code style and changed ddDrakeModel.cpp to not allocate a new `KinematicsCache` each time `setJointPositions` is called.

@patmarion can you give it a look. Let me know if there are more changes you want me to make.

PR #490, opened 2017-03-10
Fix bug in treeviewer

gizatt:
Simple tweak to get multi-axis scaling working in the treeviewer.

### DIFF
--- a/src/python/director/treeviewer.py
+++ b/src/python/director/treeviewer.py
@@ -227,7 +227,8 @@ class Geometry(object):
     def scaleGeometry(polyDataList, scale):
         if len(scale) == 1:
             scale_x = scale_y = scale_z = scale
-        elif len(scale) == 3:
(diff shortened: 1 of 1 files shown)


src/python/director/treeviewer.py
```diff
@@ -227,7 +227,7 @@ def createPolyDataFromMeshArrays(pts, faces):
     def scaleGeometry(polyDataList, scale):
         if len(scale) == 1:
             scale_x = scale_y = scale_z = scale
-        elif len(geom.float_data) == 3:
+        elif len(scale) == 3:
```
patmarion: thanks for the fix!  maybe this:

```
    else:
        assert len(scale) == 3
        scale_x, scale_y, scale_z = scale
```

gizatt: Good idea, done!

PR #495, opened 2017-04-06
Added support for obj/mtl to drake-visualizer.

liangfok:
This contains the changes to Director needed to resolve https://github.com/RobotLocomotion/drake/issues/5044.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/director/495)

### DIFF
--- a/src/python/director/drakevisualizer.py
+++ b/src/python/director/drakevisualizer.py
@@ -13,6 +13,7 @@ from director import vtkAll as vtk
 from director import vtkNumpy as vnp
 from director import visualization as vis
 from director import packagepath
+from director.fieldcontainer import FieldContainer
 
 import bot_core as lcmbot
 
@@ -109,7 +110,7 @@ class Geometry(object):
         return pd
 
     @staticmethod
-    def scaleGeometry(polyDataList, geom):
+    def scaleGeometry(meshDataList, geom):
         if len(geom.float_data) == 1:
             scale_x = scale_y = scale_z = geom.float_data[0]
         elif len(geom.float_data) == 3:
@@ -122,27 +123,30 @@ class Geometry(object):
         if scale_x != 1.0 or scale_y != 1.0 or scale_z != 1.0:
             t = vtk.vtkTransform()
             t.Scale(scale_x, scale_y, scale_z)
-            polyDataList = [filterUtils.transformPolyData(polyData, t) for polyData in polyDataList]
+            for meshData in meshDataList:
+                meshData.polyData = filterUtils.transformPolyData(meshData.polyData, t)
 
-        return polyDataList
+        return meshDataList
 
 
     @staticmethod
-    def transformGeometry(polyDataList, geom):
+    def transformGeometry(meshDataList, geom):
         t = transformUtils.transformFromPose(geom.position, geom.quaternion)
-        polyDataList = [filterUtils.transformPolyData(polyData, t) for polyData in polyDataList]
-        return polyDataList
-
+        for meshData in meshDataList:
+            meshData.polyData = filterUtils.transformPolyData(meshData.polyData, t)
+        return meshDataList
 
     @staticmethod
-    def computeNormals(polyDataList):
+    def computeNormals(meshDataList):
 
         def addNormals(polyData):
             hasNormals = polyData.GetPointData().GetNormals() is not None
             return polyData if hasNormals else filterUtils.computeNormals(polyData)
 
-        return [addNormals(polyData) for polyData in polyDataList]
+        for meshData in meshDataList:
+            meshData.polyData = addNormals(meshData.polyData)
 
+        return meshDataList
 
     @staticmethod
     def getTextureFileName(polyData):
@@ -165,7 +169,7 @@ class Geometry(object):
             imageFile = textureFileName
 
         if not os.path.isfile(imageFile):
-            print 'cannot find texture file:', textureFileName
+            print 'failed to find image file:', imageFile
             return
 
         image = ioUtils.readImage(imageFile)
@@ -196,13 +200,29 @@ class Geometry(object):
         return Geometry.PackageMap.resolveFilename(filename) or filename
 
     @staticmethod
+    def DeriveMeshVisList(geom, meshFileName, polyDataList):
+        meshDataList = []
+        for polyData in polyDataList:
+            if USE_TEXTURE_MESHES and meshFileName != "":
+                Geometry.loadTextureForMesh(polyData, meshFileName)
+            texture = Geometry.TextureCache.get(Geometry.getTextureFileName(polyData))
+            if texture:
+               color = QtGui.QColor(255, 255, 255)
+            else:
+               color = QtGui.QColor(geom.color[0]*255, geom.color[1]*255, geom.color[2]*255)
+            alpha = geom.color[3]
+            meshData = FieldContainer(polyData=polyData, color=color, alpha=alpha, texture=texture)
+            meshDataList.append(meshData)
+        return meshDataList
+
+    @staticmethod
     def loadPolyDataMeshes(geom):
 
         filename = Geometry.resolvePackageFilename(geom.string_data)
         basename, ext = os.path.splitext(filename)
 
         if filename in Geometry.MeshCache:
-            return list(Geometry.MeshCache[filename])
+            return Geometry.MeshCache[filename]
 
         preferredExtensions = ['.vtm', '.vtp', '.obj']
 
@@ -215,62 +235,62 @@ class Geometry(object):
             print 'warning, cannot find file:', filename
             return []
 
+        meshDataList = []
+
         if filename.endswith('vtm'):
             polyDataList = ioUtils.readMultiBlock(filename)
+            meshDataList = Geometry.DeriveMeshVisList(geom, filename, polyDataList)
         else:
-            polyDataList = [ioUtils.readPolyData(filename)]
+            if filename.endswith('obj'):
+                objData = ioUtils.readObjMtl(geom.string_data)
+                for polyData, actor in zip(objData.meshes, objData.actors):
+                    if objData.hasMtl:
+                        color = actor.GetProperty().GetColor()
+                        alpha = actor.GetProperty().GetOpacity()
+                        texture = actor.GetTexture()
+                    else:
+                        color = QtGui.QColor(geom.color[0]*255, geom.color[1]*255, geom.color[2]*255)
+                        alpha = geom.color[3]
+                        texture = Geometry.TextureCache.get(Geometry.getTextureFileName(polyData))
+                    meshData = FieldContainer(polyData=polyData, color=color, alpha=alpha, texture=texture)
+                    meshDataList.append(meshData)
+            else:
+                polyDataList = [ioUtils.readPolyData(filename)]
+                meshDataList = Geometry.DeriveMeshVisList(geom, filename, polyDataList)
 
-        if USE_TEXTURE_MESHES:
-            for polyData in polyDataList:
-                Geometry.loadTextureForMesh(polyData, filename)
-
-
-        Geometry.MeshCache[filename] = list(polyDataList)
-
-        return polyDataList
-
+        Geometry.MeshCache[filename] = meshDataList
+        return meshDataList
 
     @staticmethod
     def createPolyDataForGeometry(geom):
-
-        polyDataList = []
+        meshDataList = []
 
         if geom.type != lcmrl.viewer_geometry_data_t.MESH:
             polyDataList = [Geometry.createPolyDataFromPrimitive(geom)]
-
+            meshDataList = Geometry.DeriveMeshVisList(geom, "", polyDataList)
         else:
             if not geom.string_data:
                 polyDataList = [Geometry.createPolyDataFromMeshMessage(geom)]
+                meshDataList = Geometry.DeriveMeshVisList(geom, "", polyDataList)
             else:
-                polyDataList = Geometry.loadPolyDataMeshes(geom)
-                polyDataList = Geometry.scaleGeometry(polyDataList, geom)
+                meshDataList = Geometry.loadPolyDataMeshes(geom)
+                meshDataList = Geometry.scaleGeometry(meshDataList, geom)
 
-        polyDataList = Geometry.transformGeometry(polyDataList, geom)
-        polyDataList = Geometry.computeNormals(polyDataList)
+        meshDataList = Geometry.transformGeometry(meshDataList, geom)
+        meshDataList = Geometry.computeNormals(meshDataList)
 
-        return polyDataList
+        return meshDataList
 
     @staticmethod
     def createGeometry(name, geom, parentTransform):
-        polyDataList = Geometry.createPolyDataForGeometry(geom)
+        meshDataList = Geometry.createPolyDataForGeometry(geom)
+        return [Geometry(name, meshData) for meshData in meshDataList]
 
-        geometry = []
-        for polyData in polyDataList:
-            g = Geometry(name, geom, polyData, parentTransform)
-            geometry.append(g)
-        return geometry
-
-
-    def __init__(self, name, geom, polyData, parentTransform):
-        self.polyDataItem = vis.PolyDataItem(name, polyData, view=None)
-        self.polyDataItem.setProperty('Alpha', geom.color[3])
-        self.polyDataItem.actor.SetTexture(Geometry.TextureCache.get( Geometry.getTextureFileName(polyData) ))
-
-        if self.polyDataItem.actor.GetTexture():
-            self.polyDataItem.setProperty('Color', QtGui.QColor(255, 255, 255))
-
-        else:
-            self.polyDataItem.setProperty('Color', QtGui.QColor(geom.color[0]*255, geom.color[1]*255, geom.color[2]*255))
+    def __init__(self, name, meshData):
+        self.polyDataItem = vis.PolyDataItem(name, meshData.polyData, view=None)
+        self.polyDataItem.setProperty('Color', meshData.color)
+        self.polyDataItem.setProperty('Alpha', meshData.alpha)
+        self.polyDataItem.actor.SetTexture(meshData.texture)
 
         if USE_SHADOWS:
             self.polyDataItem.shadowOn()
@@ -284,7 +304,6 @@ class Link(object):
         self.geometry = []
         for g in link.geom:
             self.geometry.extend(Geometry.createGeometry(link.name + ' geometry data', g, self.transform))
-
 
     def setTransform(self, pos, quat):
         self.transform = transformUtils.transformFromPose(pos, quat)
@@ -442,7 +461,8 @@ class DrakeVisualizer(object):
                 name = linkName + ' geometry data'
                 geom = type('', (), {})
                 geom.color = [1.0,0.0,0.0,1.0]
-                g = Geometry(name, geom, polyData, link.transform)
+                meshData = FieldContainer(polyData=polyData, color=[1.0,0.0,0.0], alpha=1.0, texture=Geometry.TextureCache.get(Geometry.getTextureFileName(polyData)))
+                g = Geometry(name, meshData)
                 link.geometry.append(g)
 
                 linkFolder = self.getLinkFolder(robotNum, linkName)
--- a/src/python/director/drakevisualizer.py
+++ b/src/python/director/drakevisualizer.py
@@ -26,9 +26,6 @@ lcmrl = lcmbot
 
 from PythonQt import QtGui
 
-USE_TEXTURE_MESHES = True
-USE_SHADOWS = False
-
 class Geometry(object):
 
     MeshCache = {}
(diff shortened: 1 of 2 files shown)


patmarion:
Looks pretty good, thanks!  By the way, as I discussed last week in the slack channel, the CDash server for director is down, so CI is currently failing when it attempts to submit the test results.  I will try to get that resolved today.

I think you design might be working fine, but an alternative to consider is rather than making the logic split at `createGeometry` and `createGeometryFromObj`, you could have a single `createGeometry` function and put the logic split in the lower function `loadPolyDataMeshes`.  This function could be changed to return a list of (mesh,color,alpha,texture) tuples.

A problem with putting the split higher as you have done is that you lose some of the operations in the lower functions, like applying scaling and transforms, and the caching feature being introduced in https://github.com/RobotLocomotion/director/pull/493.

liangfok:
Thanks for the review! I will refactor as suggested.

patmarion:
For non-obj meshes, they could return None for color and alpha, in which case the values from the geometry message are used.  For obj with mtl, you have to decide whether you want to use the color from the .mtl file, or the color from the URDF (geometry message).

patmarion:
Also, as currently implemented, `ioUtils.readObjMtl()` does not make it obvious whether the returned actor's color was specified from a .mtl file, or whether the actor color is just the default color (perhaps because there was no .mtl file associated with the .obj file)

Lucy-tri:
Glad to see this.  It sounds like the Drake doc (http://drake.mit.edu/models.html#texture-file-format) will have to be updated with this change. (That doc describes drake-visualizer too).  Will drake-visualizer continue to support  VTM/VTP texture files, as it does now?

patmarion:
Yes, vtm files should continue to be supported.  I think this feature as implemented is currently breaking support for some models, because it now preferences obj files over vtm.  But we have several models that have vtm files generated from dae+textures, and obj without mtl, and so the vtm files should be preferred.

liangfok:
@patmarion, please check `r3` of this PR.  I still have to polish it some more but wanted to check whether it's going in the right direction. Thanks!

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/robotlocomotion/director/495#-:-Kh8cf4XXp14PuGMLTtU:b-j3zf61)*

patmarion:
definitely headed in the right direction, but the implementation has gotten really messy.  I don't think the functions should have multiple list return values.  It's time to return a list of structs.  Director has a very convenient class for this.  Try something like:


```
from director.fieldcontainer import FieldContainer

...


result = []

# load mesh and color and texture data
for ...:
    f = FieldContainer(polyData=polyData, color=color, alpha=alpha, texture=texture)
    result.append(f)

return result

```

Now you can access the fields by name:

```
f = result[0]
f.polyData
f.color
f.texture
```

liangfok:
Thanks @patmarion. In `r5` I switched to use `FieldContainer` as suggested though I believe it still needs quite a bit of polishing. PTAL?

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/robotlocomotion/director/495#-:-Kh9D_LAZTBToU4BFs66:b-kfofgz)*

liangfok:
I completed a round of polishing in `r6`.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/robotlocomotion/director/495#-:-KhAEfVSu8Kd1MlFdfpC:b18ujt9)*

patmarion:
i read through it, thanks for working on this!  I will leave some in-line comments.  Overall I think there are some simplifications that could be made.  The existing logic was already messy due to its history, and this patch makes it further complicated.  I might take a pass over this, if i have time, but we might be able to merge it sooner than that, after you get a chance to address a few more review comments.  Thanks!

src/python/director/drakevisualizer.py
```diff
@@ -196,13 +202,32 @@ def resolvePackageFilename(filename):
         return Geometry.PackageMap.resolveFilename(filename) or filename
 
     @staticmethod
+    def DeriveMeshVisList(geom, meshFileName, polyDataList):
```
patmarion: style: camelCase method name please

src/python/director/drakevisualizer.py
```diff
@@ -215,62 +240,68 @@ def loadPolyDataMeshes(geom):
...
-
-        if USE_TEXTURE_MESHES:
-            for polyData in polyDataList:
-                Geometry.loadTextureForMesh(polyData, filename)
-
-
-        Geometry.MeshCache[filename] = list(polyDataList)
-
-        return polyDataList
+            if os.path.isfile(basename + '.mtl'):
```
patmarion: bug: this is not the correct way to determine whether an mtl file exists.  Also, we probably want all obj files to be loaded with the `ioUtils.readObjMtl` function, not just obj files with `.mtl` material files.

src/python/director/drakevisualizer.py
```diff
@@ -215,62 +240,68 @@ def loadPolyDataMeshes(geom):
...
+                    color = actor.GetProperty().GetColor()
+                    alpha = actor.GetProperty().GetOpacity()
+                    texture = actor.GetTexture()
+                    mesh_vis = FieldContainer( \
+                        polyData = polyData, color = color, alpha = alpha, \
+                        texture = texture)
+                    mesh_vis_list.append(mesh_vis);
+            else:
+                polyDataList = [ioUtils.readPolyData(filename)]
+                mesh_vis_list = Geometry.DeriveMeshVisList(geom = geom, \
```
patmarion: style: don't use `name=arg` style arg list, just a regular positional arg list.  this comment applies to other areas in the diff, too

src/python/director/drakevisualizer.py
```diff
@@ -215,62 +240,68 @@ def loadPolyDataMeshes(geom):
...
     @staticmethod
     def createGeometry(name, geom, parentTransform):
-        polyDataList = Geometry.createPolyDataForGeometry(geom)
-
+        mesh_vis_list = Geometry.createPolyDataForGeometry(geom)
         geometry = []
-        for polyData in polyDataList:
-            g = Geometry(name, geom, polyData, parentTransform)
+        for mesh_vis in mesh_vis_list:
+            g = Geometry(name, mesh_vis.polyData, mesh_vis.color, \
```
patmarion: maybe just pass the struct to the ctor instead of the individual of fields?

src/python/director/drakevisualizer.py
```diff
@@ -215,62 +240,68 @@ def loadPolyDataMeshes(geom):
...
 
     @staticmethod
     def createGeometry(name, geom, parentTransform):
-        polyDataList = Geometry.createPolyDataForGeometry(geom)
-
+        mesh_vis_list = Geometry.createPolyDataForGeometry(geom)
         geometry = []
-        for polyData in polyDataList:
-            g = Geometry(name, geom, polyData, parentTransform)
+        for mesh_vis in mesh_vis_list:
```
patmarion: loop could be a list comprehension

liangfok:
Thanks for the review! I will address your comments ASAP.

liangfok:
@patmarion, I addressed your comments. PTAL.

---

Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/python/director/drakevisualizer.py, line 113 at r6](https://reviewable.io:443/reviews/robotlocomotion/director/495#-Kh2Zer0keu56iRzFVhp-r6-113:-KhOkSIMOHc0S1b5MjTq:b-896fix) ([raw file](https://github.com/robotlocomotion/director/blob/9578b201e64cbdef6c851e65092bcaac8568d2ee/src/python/director/drakevisualizer.py#L113)):*
<details><summary><i>Previously, patmarion (Pat Marion) wrote…</i></summary><blockquote>

style: please use camelCase, not underscores.  Maybe we can name this variable `meshDataList`?  this comment applies for the whole diff.
</blockquote></details>

Done.

---

*[src/python/director/drakevisualizer.py, line 127 at r6](https://reviewable.io:443/reviews/robotlocomotion/director/495#-Kh2Zer0keu56iRzFVhp-r6-127:-KhOqn-c5AHhlY1qOsh7:b-896fix) ([raw file](https://github.com/robotlocomotion/director/blob/9578b201e64cbdef6c851e65092bcaac8568d2ee/src/python/director/drakevisualizer.py#L127)):*
<details><summary><i>Previously, patmarion (Pat Marion) wrote…</i></summary><blockquote>

style: please don't wrap lines with `\`.  I don't enforce line length limits in these files.  this comment applies for the whole diff.
</blockquote></details>

Done.

---

*[src/python/director/drakevisualizer.py, line 205 at r6](https://reviewable.io:443/reviews/robotlocomotion/director/495#-Kh2Zer0keu56iRzFVhp-r6-205:-KhOkdz2Tkg6Z9NYhz1a:b-896fix) ([raw file](https://github.com/robotlocomotion/director/blob/9578b201e64cbdef6c851e65092bcaac8568d2ee/src/python/director/drakevisualizer.py#L205)):*
<details><summary><i>Previously, patmarion (Pat Marion) wrote…</i></summary><blockquote>

style: camelCase method name please
</blockquote></details>

Done.

---

*[src/python/director/drakevisualizer.py, line 250 at r6](https://reviewable.io:443/reviews/robotlocomotion/director/495#-Kh2Zer0keu56iRzFVhp-r6-250:-KhOr1gfUZdwtlVYipWz:b-896fix) ([raw file](https://github.com/robotlocomotion/director/blob/9578b201e64cbdef6c851e65092bcaac8568d2ee/src/python/director/drakevisualizer.py#L250)):*
<details><summary><i>Previously, patmarion (Pat Marion) wrote…</i></summary><blockquote>

bug: this is not the correct way to determine whether an mtl file exists.  Also, we probably want all obj files to be loaded with the `ioUtils.readObjMtl` function, not just obj files with `.mtl` material files.
</blockquote></details>

Done.

---

*[src/python/director/drakevisualizer.py, line 262 at r6](https://reviewable.io:443/reviews/robotlocomotion/director/495#-Kh2Zer0keu56iRzFVhp-r6-262:-KhOlBfR_voc_ktwG2gG:b-896fix) ([raw file](https://github.com/robotlocomotion/director/blob/9578b201e64cbdef6c851e65092bcaac8568d2ee/src/python/director/drakevisualizer.py#L262)):*
<details><summary><i>Previously, patmarion (Pat Marion) wrote…</i></summary><blockquote>

style: don't use `name=arg` style arg list, just a regular positional arg list.  this comment applies to other areas in the diff, too
</blockquote></details>

Done.

---

*[src/python/director/drakevisualizer.py, line 294 at r6](https://reviewable.io:443/reviews/robotlocomotion/director/495#-Kh2Zer0keu56iRzFVhp-r6-294:-KhOqhFsKpvRQrK9W2gF:b-896fix) ([raw file](https://github.com/robotlocomotion/director/blob/9578b201e64cbdef6c851e65092bcaac8568d2ee/src/python/director/drakevisualizer.py#L294)):*
<details><summary><i>Previously, patmarion (Pat Marion) wrote…</i></summary><blockquote>

loop could be a list comprehension
</blockquote></details>

Done.

---

*[src/python/director/drakevisualizer.py, line 295 at r6](https://reviewable.io:443/reviews/robotlocomotion/director/495#-Kh2Zer0keu56iRzFVhp-r6-295:-KhOokYZsKbqkUC1biMd:b-896fix) ([raw file](https://github.com/robotlocomotion/director/blob/9578b201e64cbdef6c851e65092bcaac8568d2ee/src/python/director/drakevisualizer.py#L295)):*
<details><summary><i>Previously, patmarion (Pat Marion) wrote…</i></summary><blockquote>

maybe just pass the struct to the ctor instead of the individual of fields?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/robotlocomotion/director/495#-:-KhP4wkeeJCsaywqM4V4:b-d5d1qi)*

src/python/director/drakevisualizer.py
```diff
@@ -196,13 +200,29 @@ def resolvePackageFilename(filename):
...
+            if USE_TEXTURE_MESHES and meshFileName != "":
+                Geometry.loadTextureForMesh(polyData, meshFileName)
+            texture = Geometry.TextureCache.get(Geometry.getTextureFileName(polyData))
+            if texture:
+               color = QtGui.QColor(255, 255, 255)
+            else:
+               color = QtGui.QColor(geom.color[0]*255, geom.color[1]*255, geom.color[2]*255)
+            alpha = geom.color[3]
+            meshData = FieldContainer(polyData=polyData, color=color, alpha=alpha, texture=texture)
+            meshDataList.append(meshData);
```
patmarion: remove semi-colon

src/python/director/drakevisualizer.py
```diff
@@ -215,62 +235,57 @@ def loadPolyDataMeshes(geom):
...
-
-        return polyDataList
+            if filename.endswith('obj'):
+                polyDataList, actorList = ioUtils.readObjMtl(geom.string_data)
+                for polyData, actor in zip(polyDataList, actorList):
+                    color = actor.GetProperty().GetColor()
+                    alpha = actor.GetProperty().GetOpacity()
+                    texture = actor.GetTexture()
+                    meshData = FieldContainer(polyData = polyData, color = color, alpha = alpha, texture = texture)
+                    meshDataList.append(meshData);
```
patmarion: remove semi-colon

src/python/director/drakevisualizer.py
```diff
@@ -215,62 +235,57 @@ def loadPolyDataMeshes(geom):
...
-        Geometry.MeshCache[filename] = list(polyDataList)
-
-        return polyDataList
+            if filename.endswith('obj'):
+                polyDataList, actorList = ioUtils.readObjMtl(geom.string_data)
+                for polyData, actor in zip(polyDataList, actorList):
+                    color = actor.GetProperty().GetColor()
+                    alpha = actor.GetProperty().GetOpacity()
+                    texture = actor.GetTexture()
+                    meshData = FieldContainer(polyData = polyData, color = color, alpha = alpha, texture = texture)
```
patmarion: remove spaces around `=`

src/python/director/drakevisualizer.py
```diff
@@ -215,62 +235,57 @@ def loadPolyDataMeshes(geom):
...
-
-        if USE_TEXTURE_MESHES:
-            for polyData in polyDataList:
-                Geometry.loadTextureForMesh(polyData, filename)
-
-
-        Geometry.MeshCache[filename] = list(polyDataList)
-
-        return polyDataList
+            if filename.endswith('obj'):
```
patmarion: i think i need to test this, but i think there is one more remaining bug, which is the case of urdf files that specify visuals with color and mesh obj files, and those obj files do not have associated mtl files.  In this case, the color specified in the urdf should be used.  I think we have urdf files like this.  But as implemented here, the urdf color will be ignored.

patmarion:
thanks for working on this further! sorry for dragging out the review, thanks for your patience.

liangfok:
Thank YOU for all of the rounds of reviews! No worries about the latency -- I'm always happy to learn how to improve the code.

---

Review status: 0 of 1 files reviewed at latest revision, 11 unresolved discussions.

---

*[src/python/director/drakevisualizer.py, line 215 at r8](https://reviewable.io:443/reviews/robotlocomotion/director/495#-Kh2Zer0keu56iRzFVhp-r8-215:-KhPTST8sEjMctCD3ix9:b-896fix) ([raw file](https://github.com/robotlocomotion/director/blob/363840635c0002c7e7437fd9417907fcdc3d73da/src/python/director/drakevisualizer.py#L215)):*
<details><summary><i>Previously, patmarion (Pat Marion) wrote…</i></summary><blockquote>

remove semi-colon
</blockquote></details>

Done.

---

*[src/python/director/drakevisualizer.py, line 244 at r8](https://reviewable.io:443/reviews/robotlocomotion/director/495#-Kh2Zer0keu56iRzFVhp-r8-244:-KhPTUo9V2Zo-x7PEPH0:b-kd0clh) ([raw file](https://github.com/robotlocomotion/director/blob/363840635c0002c7e7437fd9417907fcdc3d73da/src/python/director/drakevisualizer.py#L244)):*
<details><summary><i>Previously, patmarion (Pat Marion) wrote…</i></summary><blockquote>

i think i need to test this, but i think there is one more remaining bug, which is the case of urdf files that specify visuals with color and mesh obj files, and those obj files do not have associated mtl files.  In this case, the color specified in the urdf should be used.  I think we have urdf files like this.  But as implemented here, the urdf color will be ignored.
</blockquote></details>

Good point.  I will also test this ASAP.

---

*[src/python/director/drakevisualizer.py, line 250 at r8](https://reviewable.io:443/reviews/robotlocomotion/director/495#-Kh2Zer0keu56iRzFVhp-r8-250:-KhPTPJI-Pq8zQzB_HPc:b-896fix) ([raw file](https://github.com/robotlocomotion/director/blob/363840635c0002c7e7437fd9417907fcdc3d73da/src/python/director/drakevisualizer.py#L250)):*
<details><summary><i>Previously, patmarion (Pat Marion) wrote…</i></summary><blockquote>

remove spaces around `=`
</blockquote></details>

Done.

---

*[src/python/director/drakevisualizer.py, line 251 at r8](https://reviewable.io:443/reviews/robotlocomotion/director/495#-Kh2Zer0keu56iRzFVhp-r8-251:-KhPTPxF6aGzOkYb6ZBB:b-896fix) ([raw file](https://github.com/robotlocomotion/director/blob/363840635c0002c7e7437fd9417907fcdc3d73da/src/python/director/drakevisualizer.py#L251)):*
<details><summary><i>Previously, patmarion (Pat Marion) wrote…</i></summary><blockquote>

remove semi-colon
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/robotlocomotion/director/495#-:-KhPTYTNLnMIvInq0WUE:b-tvs5pk)*

src/python/director/drakevisualizer.py
liangfok: I confirm that this is a bug. It can be replicated by running the `controlled_iiwa_world_demo`, which uses `/manipulation/models/iiwa_description/urdf/iiwa14_polytope_collision.urdf`. This model references .obj files that do not have associated .mtl files and should be colored based on what the URDF specifies. I will fix this ASAP.

liangfok: Done. I tested it using the aforementioned `controlled_iiwa_world_demo`. PTAL.

Lucy-tri:
Don't forget to update the Drake doc. Good to see this.

---

Review status: 0 of 2 files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/robotlocomotion/director/495#-:-KhSHQ7XrlDWkQv4vfrv:bgvyh3d)*

liangfok:
Thanks for the reminder. I'll be sure to update Drake's Doc in the same PR that bumps Director's version in Drake.

---

Review status: 0 of 2 files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/robotlocomotion/director/495#-:-KhSHkORSAXjGm897sLo:b-ex8b01)*

patmarion:
I think this looks good, thanks!  I would like to do some manual testing myself prior to merging, since there isn't any good test coverage for a lot of these cases.  Hopefully I'll find some time to do that today and then I'll merge it.

liangfok:
Wonderful! Let me know if there are any additional fixes I can apply.

Lucy-tri:
Is there any documentation for drake-visualizer that might mention what files it supports (or a banner comment in drake-visualizer, or something)? 

---

Review status: 0 of 2 files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/robotlocomotion/director/495#-:-KhSXmRs3FA2P-Sr_vVh:bn6oka1)*

patmarion:
Docs are very sparse.  User guide information [goes here](https://github.com/RobotLocomotion/director/tree/master/docs/sphinx) and there's always room to add high quality comments to the .py files.

Lucy-tri:
So if it makes sense, it would be great to have a sentence or two  somewhere, for users of drake-visualizer, especially now that versions matter for what files you want to use.  Maybe if the help for drake-visualizer  could print out what files it supports (I'm thinking especially of the non-programmers who use it)?  Or at least a banner comment in drake-visualizer.py.  So when you run drake-visualizer  with no args, it prints help.   Again, really glad to see this PR.

---

Review status: 0 of 2 files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/robotlocomotion/director/495#-:-KhSejkmkMpv2PkeOaTy:b-rmmn6p)*

liangfok:
Drake users will probably not use `drakevisualizer.py` directly. Instead, they'll use a C++ executable called `drake-visualizer` that is a wrapper around the Python script (it's implemented in `director/src/app/drakeVisualizerApp.cpp`). Because of this, I'm hesitant to place the documentation within `drakevisualizer.py`, and will probably include it in Drake's `drake/doc/models.rst` document.

---

Review status: 0 of 2 files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/robotlocomotion/director/495#-:-KhTUyAzO5VY4RAKm-zP:b-xom85a)*

PR #499, opened 2017-04-06
Fix vtk packaging

patmarion:


patmarion:
@jwnimmer-tri if you're interested please try:

http://people.csail.mit.edu/patmarion/software/director/releases/director-0.1.0-73-gd931e35-ubuntu-16.04.tar.gz

This binary has vtk5 included in the package so that it does not depend on system vtk.  I verified it works on my ubuntu 16 installation when:

  - no version of vtk is system installed
  - vtk6 is system installed

It should work without requiring any custom PYTHONPATH and LD_LIBRARY_PATH.  Just extract and run bin/drake-visualizer.  As before, it still depends on system python27 and qt4.

jwnimmer-tri:
Thanks, very exciting!  It'll probably be the weekend when I get a moment to check.

patmarion:
Note, I created the package manually.  The director Dockerfile is designed for running on Travis-CI.  I'd be happy to also create Dockerfiles designed to be run anywhere, and produce packages with vtk on ubuntu 14 and 16. That would be straightforward.  Then you can run those Dockerfiles on any server you'd like in order to automatically produce packages, without relying on Travis-CI.

PR #500, opened 2017-04-06
Add non-conservative set of dependencies for Ubuntu 16.04

EricCousineau-TRI:
This adds a non-conservative set of dependencies for installing on Ubuntu 16.04 for a development machine.

Drawn from `oh-distro`:
https://github.com/openhumanoids/oh-distro/blob/eeee1d8/README.rst#dependencies

I have added this to `drake-distro`'s pods install script as a very rough start, but please let me know if these prereqs should be listed elsewhere.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/director/500)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -84,6 +84,12 @@ The dependencies can be installed on Ubuntu using apt-get:
 
     sudo apt-get install cmake libvtk5-qt4-dev libeigen3-dev python-dev python-vtk python-numpy
 
+If you are running Ubuntu 16.04, please install the following non-conservative set of
+dependencies by executing:
+
+::
+
+    ./distro/pods/drake-distro/install_prereqs.sh ubuntu_16_04
 
 Building
 ========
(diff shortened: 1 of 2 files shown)


patmarion:
sorry, I don't want to add all of these.  I know the intent is to be "non-conservative", but this is too non-conservative, most of those are not required dependencies.  Was the only issue that you did not have boost development libraries for PCL?  Instead, maybe you could just update the README to say something like:

Additional dependencies:
  If you plan you build director with extra options enabled, such as Point Cloud Library, please refer to the documentation of those projects for required dependencies.

patmarion:
Also, `director/distro/pods/drake-distro` is actually unused now and the whole directory should be removed at some point.

EricCousineau-TRI:
Makes sense!
In this case, I will try to scope the setup steps solely for working with the IIWA, and move the `README` there, as my current goal is to smooth out the IIWA setup process using `director`.

PR #506, opened 2017-04-26
Try to fix docs build

rdeits:
@patmarion this PR is just here to get Travis to build the docs.


patmarion:
thanks and good luck!

rdeits:
I got it! It looks like we needed an upgraded version of `breathe`. 

I also cleaned up some things in the doxyfile that were giving deprecation warnings.

patmarion:
hooray!  good stuff.

so here's how the process works:  travis-ci decrypts an ssh key and uses that to scp the packages and docs to a dropbox folder on drc001 which is on my desk in 32-380.  When the files show up in my dropbox i manually upload them to the web.  would be nice to just have travis upload directly instead of staging thing in my dropbox.

PR #514, opened 2017-05-29
Add option to automatically remove stale disparity data

wxmerkt:
Adds two new properties to DisparityPointCloudItem to allow data to expire and be automatically removed if no new messages have been received for a given timeout/expiration time.

### DIFF
--- a/src/python/director/segmentation.py
+++ b/src/python/director/segmentation.py
@@ -180,6 +180,9 @@ class DisparityPointCloudItem(vis.PolyDataItem):
         self.imageManager = imageManager
         self.cameraName = cameraName
         self.setProperty('Visible', False)
+        self.addProperty('Remove Stale Data', False)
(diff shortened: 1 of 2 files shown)


src/python/director/segmentation.py
```diff
@@ -199,12 +202,13 @@ def onRemoveFromObjectModel(self):
         self.timer.stop()
 
     def update(self):
-
         utime = self.imageManager.queue.getCurrentImageTime(self.cameraName)
         if utime == self.lastUtime:
+            if self.getProperty('Remove Stale Data') and ((time.time()-self.lastDataReceivedTime) > self.getProperty('Stale Data Timeout')):
+                self.setPolyData(vtk.vtkPolyData())
```
patmarion: maybe also add a check for self.polyData.GetNumberOfPoints() > 0 so that it does not continuously reset the polyData?  Not sure there's really anything wrong with it as written tho.

wxmerkt:
Thanks for the review and comment - I've added the additional check

PR #515, opened 2017-06-06
update WrapPythonQt.py

patmarion:
Adds --auto-class-includes option to determine whether or not the wrapper
should automatically guess the header given a class name.  This option
works well for Director source files but other projects may have different
naming conventions.


jcfr:
I guess we both came up with a similar scripts ... see https://github.com/commontk/CTK/blob/06ac6eb5908981a6fc63596aa8f9c153733ba13d/CMake/ctkWrapPythonQt.py

when I have a chance, I will try to unify both approach and add the script directly to PythonQt repo.

patmarion:
Yup. I used this script for several projects at Kitware, for example, here's another copy: https://github.com/Kitware/ParaView/blob/master/Plugins/PythonQtPlugin/WrapPythonQt.py

PR #518, opened 2017-06-12
Port vtk7 qt5

patmarion:


jamiesnape:
So this will no longer work with Drake since all the packages now refer to VTK 7.1. FYI @stonier.

sankhesh:
@patmarion Just to confirm, travis-ci is using the OpenGL backend on this PR, right?

patmarion:
> So this will no longer work with Drake since all the packages now refer to VTK 7.1. FYI @stonier.

I think this should work with drake.  In the drake jenkins builds you can install the vtk8 rc and everything should work.   If you use the director superbuild just compile with `-DUSE_SYSTEM_VTK=ON` and `-DVTK_DIR=<vtk install location>`.   I am using VTK 7.1 on Travis CI.

The only difference between this and #508 is this does not have the `QVTKOpenGLWidget` integration.

jamiesnape:
Find package will definitely fail on Mac for Drake.

patmarion:
> @patmarion Just to confirm, travis-ci is using the OpenGL backend on this PR, right?

Correct, it is testing with VTK 7.1 compiled with OpenGL backend on mac and ubuntu.  It is currently only testing Qt4 but I intend to add Qt5 testing to travis, too.

patmarion:
> Find package will definitely fail on Mac for Drake.

Use `-DVTK_DIR=...` or set `CMAKE_PREFIX_PATH` if you have a non-standard location.

jamiesnape:
> Use -DVTK_DIR=... or set CMAKE_PREFIX_PATH if you have a non-standard location.

Obviously. We should not have to for Drake. `/usr/local/opt/vtk@8.0` etc. is the standard location.

patmarion:
VTK 8 is not officially released and your Homebrew formula is keg only.  So I think passing `-DVTK_DIR` is a reasonable solution.  I can also add a hint path the the director superbuild if it would really help.

sankhesh:
> I can also add a hint path the the director superbuild if it would really help

If [this](https://github.com/RobotLocomotion/director/pull/508/commits/ab0388f09bc635b4c8b296f6f1a824ef9f040564#diff-575c577d7999df9a68e994e3c3c15b78R373) change is not overwritten (from #508), the hint exists.

stonier:
First request, make it much simpler and bulletproofed when requesting a build with drake:

If `USE_DRAKE:=ON`, 

* call `find_package(DRAKE)
* on linux set HINTS to find VTK at drake's external path
* on apple set HINTS to find TK at the homebrew install path
* use these to call `find_package(VTK REQUIRED HINTS ...)` appropriately
* check the VTK version found, if not 8.0, fail fast with a warning

This will mean making sure the VTK logic should shift to post-drake logic. @jamiesnape correct me if this doesn't make sense or pat needs some extra clarity.

jamiesnape:
Sounds correct to me.

stonier:
Second problem is that any launching will fail hard because the `PYTHONPATH` will not be pointing to VTK, nor Drake's directories. The user will need to prefix to it as per [drake_visualiser_linux.sh](https://github.com/RobotLocomotion/drake/blob/master/tools/drake_visualizer_linux.sh) and [drake_visualiser_apple.sh](https://github.com/RobotLocomotion/drake/blob/master/tools/drake_visualizer_apple.sh).

Not really something to hold up this script though - we can do something over in spartan.

stonier:
Shifted request to #520.

PR #522, opened 2017-06-15
Add simple slider animation when playing. Enable playback from non-zero.

EricCousineau-TRI:
This is a basic change to (a) enable playback from a non-zero start value and (b) show the slider animated during playback.

Motivation:
I am using `sim_playback` to record a simulation with different actors enabled / disabled per scene. If it starts from 0.0, then it will call the `LOAD` messages, which will cause the visibilities to all be turned back on.
(Animation was just so I can see where I was at.)

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/director/522)

### DIFF
--- a/src/python/director/lcmlogplayer.py
+++ b/src/python/director/lcmlogplayer.py
@@ -149,41 +149,59 @@ class LcmLogPlayerGui(object):
         stopButton = QtGui.QPushButton('Stop')
         slider = QtGui.QSlider(QtCore.Qt.Horizontal)
         slider.maximum = int(logPlayer.getEndTime()*100)
+        text = QtGui.QLineEdit()
+        text.text = '0.0'
         playButton.connect('clicked()', self.onPlay)
         stopButton.connect('clicked()', self.onStop)
         slider.connect('valueChanged(int)', self.onSlider)
+        text.connect('returnPressed()', self.onText)
 
         l = QtGui.QHBoxLayout(w)
         l.addWidget(slider)
+        l.addWidget(text)
         l.addWidget(playButton)
         l.addWidget(stopButton)
 
         self.slider = slider
+        self.text = text
         self.widget = w
         self.widget.show()
 
-    def _getTime(self, value=None):
+    def _getSliderTime(self, value=None):
         if value is None:
             value = self.slider.value
         t = self.logPlayer.getEndTime()*value/self.slider.maximum
         return t
 
-    def _getValue(self, t):
+    def _getSliderValue(self, t):
         value = t / self.logPlayer.getEndTime() * self.slider.maximum
         return int(round(value))
 
+    def _updateTime(self, t):
+        with BlockSignals(self.slider, self.text):
+            self.slider.value = self._getSliderValue(t)
+            self.text.text = str(t)
+
     def onPlay(self):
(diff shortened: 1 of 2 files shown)


src/python/director/lcmlogplayer.py
```diff
@@ -142,29 +149,71 @@ def __init__(self, logPlayer):
...
+
+    def onText(self, *args):
+        try:
+            t = float(self.text.text)
+            self.skipTo(t)
+        except ValueError:
+            pass
+
+# @ref https://stackoverflow.com/a/35000974/7829525
+class BlockSignals(object):
```
patmarion: awesome, can you put this class into `qtutils.py`

https://github.com/RobotLocomotion/director/blob/master/src/python/director/qtutils.py

patmarion:
looks good to me!  Just had one comment about the utility class.  Let me know if you want to make that change, then Ill merge.

Can you tell me where you are using director?  Director master is now upgraded to newer vtk, so it is not compatible with spartan.  Spartan actually uses a dev branch of director, I think, so if you use this with spartan you'll need to add this commit to that branch.  If you are using this in drake, i think you have to wait for new drake-visualizer binaries.

patmarion:
Thanks!  Are you finished adding features on this branch or do you have more coming?  I think it's pretty good as written, but since it's a larger change I'd like to suggest some possible design tweaks.  I'm a little bit busy today though, are you blocked waiting for this merge or are you ok using your branch for today?

If I want to give it a try, are you using this with a modified sim_playback in spartan?  Also, any movies you created, I'd like to see!

EricCousineau-TRI:
Sorry for not posting an update after pushing!

I am using `director` in a side `dev` branch of `drake`, just to record a few videos from `drake` (not really using `spartan`, was just doing so to test it outside of my dev branch), and I will be migrating the useful portions of my changes back to both `master`s.
I can definitely cherry-pick this commit range onto the dev branch within `spartan` when I have a chance.

Can't say that the video is anything super impressive, but I'll email it to ya.
I just wanted to fade in / out a few objects, and realized I couldn't do that with non-deterministic frames being recorded between each playback, hence the mods. (I wanted the same timestamps to be snapshot'd to do some basic blending in Adobe Premier with synchronized frames.)

And yeah, some of the synchronized stuff is definitely hacky, so I can back that out, and just stay on my own branch.
Only caveat is, I probably won't be able to clean it up in the near future (and may not need the feature for the next couple of weeks), but I'd definitely like to get your suggestions.
I'll separate those changes from this PR, and put those elsewhere. (I think I had meant not to push the sync stuff on this one, so sorry about the extra noise!)

I guess one thing I'd like to do as well is move your script to a default-loaded widget, as it's super useful, so maybe I could incorporate your suggestions in that PR.

patmarion:
Yeah you can definitely make it a permanent widget in `drake-visualizer`.  Just add some code into this function here:

https://github.com/RobotLocomotion/director/blob/master/src/python/director/mainwindowapp.py#L419

You can make it a docked panel, or keep it as a floating top level widget.  If you want it as a floating widget, default hidden but accessible from the `View` menu, I think you can add code like:

```
    fields.app.addWidgetToViewMenu(widget)
    widget.hide()
```

PR #523, opened 2017-06-16
Integrate QVTKOpenGLWidget from VTK 8

sankhesh:
QVTKOpenGLWidget changes ported from #508

### DIFF
--- a/distro/superbuild/cmake/externals.cmake
+++ b/distro/superbuild/cmake/externals.cmake
@@ -367,7 +367,7 @@ else()
 
   ExternalProject_Add(vtk
     GIT_REPOSITORY git://vtk.org/VTK.git
-    GIT_TAG v7.1.1
+    GIT_TAG 28deb5620e56f535fc92ff8c73ef00f54923839b
     CMAKE_CACHE_ARGS
       ${default_cmake_args}
       ${python_args}
@@ -375,7 +375,7 @@ else()
       -DBUILD_SHARED_LIBS:BOOL=ON
       -DBUILD_TESTING:BOOL=OFF
       -DBUILD_EXAMPLES:BOOL=OFF
-      -DVTK_RENDERING_BACKEND:STRING=OpenGL
+      -DVTK_RENDERING_BACKEND:STRING=OpenGL2
       -DVTK_QT_VERSION:STRING=${DD_QT_VERSION}
       -DVTK_PYTHON_VERSION=2
       -DModule_vtkGUISupportQt:BOOL=ON
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -34,8 +34,19 @@ include_directories(${PYTHON_INCLUDE_DIRS})
 
 set(decorator_file ${CMAKE_CURRENT_BINARY_DIR}/ddPythonQtDecorators.h)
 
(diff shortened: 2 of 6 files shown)


patmarion:
Hi Sankhesh,

I am very excited to have this code in master, but I withheld it from #518 because I think there are a few more cleanups required before it's ready:

1) Need to enable Qt5 and vtk8 testing on CI

2) Should the `QVTKOpenGLWidget` widget only be used for qt5 and vtk8 with OpenGL2 backend?  Current the code only checks the qt version.  We could put the version checking into the cmake code and use a configured header to explicitly say whether or not to use `QVTKOpenGLWidget`, rather than doing the repeating the version check in all the places in the c++ code.

3) The initialization that happens before QApplication could be moved into a reusable class and called from all the locations that construct QApplication.  CI tests use `directorPython` from `consoleApp.cpp`, but `drakeVisualizerApp.cpp`.

jamiesnape:
FYI @stonier. We need this pretty urgently for Drake.

patmarion:
I thought you built the current set of binaries from the #508 branch instead of master.  Can you do that for the next round of packages so that you aren't blocked by this?

sankhesh:
Hi Pat, 

>1. Need to enable Qt5 and vtk8 testing on CI

@jamiesnape has some ideas for getting `xvfb` to open an OpenGL 3.2 core context.

> 2.  Should the QVTKOpenGLWidget widget only be used for qt5 and vtk8 with OpenGL2 backend? 

Yes, the `QVTKOpenGLWidget would only be available for the OpenGL2 backend. See https://gitlab.kitware.com/vtk/vtk/blob/72296e8d8f43fad32f08afefd81d471c4d8548b5/GUISupport/Qt/CMakeLists.txt#L37-41

> Current the code only checks the qt version. We could put the version checking into the cmake code and use a configured header to explicitly say whether or not to use QVTKOpenGLWidget, rather than doing the repeating the version check in all the places in the c++ code.

That's a good idea. It would make the C++ code much cleaner. Although, would you mind handling it as part of the CMake code cleanup that you proposed during our conversation?

> The initialization that happens before QApplication could be moved into a reusable class and called from all the locations that construct QApplication. CI tests use directorPython from consoleApp.cpp, but drakeVisualizerApp.cpp.

Sure. It could be. But since its just two consumers (and a three lines of code), we could do it as part of a separate follow-on PR. What do you think?

jamiesnape:
We want to switch off VTK 5 support, so the CMake build including Director needs VTK 8 support. At present there are rendering issues without this.

jamiesnape:
> @jamiesnape has some ideas for getting xvfb to open an OpenGL 3.2 core context.

`sudo apt-get install libgl1-mesa-glx-lts-xenial` if I think.

sankhesh:
> `sudo apt-get install libgl1-mesa-glx-lts-xenial`, I think.

I thought Travis-ci is already compiling against mesa (or I think I tried it). I'll try it again to be sure.

jamiesnape:
Not the Xenial one.

patmarion:
Of the 3 cleanups I listed, only 2) and 3) are important code changes.  1) is for CI testing, but I don't mind delaying that.  I think 2) and 3) are quick to implement.

sankhesh:
@jamiesnape @patmarion #528 supersedes this PR.

PR #525, opened 2017-06-16
[WIP] Add basic synchronized video frame recording, for use with LcmLogPlayer

EricCousineau-TRI:
Following comment in PR #522, this is a preliminary addition to start enabling recording deterministic videos (at least as far as I'm aware - please let me know if there's a better way!) such that a video should always have the same number of frames at the same time instants - with this current (hacky) PR, this is synchronized with LCM timestamps, such that you should effectively always produce the exact same video from an LCM Log.

An example application is blending between two videos with certain objects hidden, and not having to do any video synchronization.

My initial guess on improvements would be to separate the `SreenGrabberPanel` recording functionality in its own class, and maybe have that share some common Qt signal with the GUI elements, such that these singleton hacks aren't (as) necessary.

\cc @patmarion @sammy-tri

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/director/525)

### DIFF
--- a/src/python/director/screengrabberpanel.py
+++ b/src/python/director/screengrabberpanel.py
@@ -27,9 +27,11 @@ class WidgetDict(object):
 
 
 class ScreenGrabberPanel(object):
+    # HACK
+    instance = None
 
     def __init__(self, view):
-
+        ScreenGrabberPanel.instance = self
         self.view = view
 
         loader = QtUiTools.QUiLoader()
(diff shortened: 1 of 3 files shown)


src/python/director/screengrabberpanel.py
```diff
@@ -27,9 +27,11 @@ def __init__(self, widgets):
 
 
 class ScreenGrabberPanel(object):
+    # HACK
+    instance = None
 
     def __init__(self, view):
-
+        ScreenGrabberPanel.instance = self
```
patmarion: we should remove this.  I can help show you how to get the instance without adding this variable.  I might propose that LCMLogPlayer fires off callbacks, and we add a new class called something like `LcmLogVideoRecorder` which listens to the callbacks and has a ref to the screen grabber panel to record movies.

EricCousineau-TRI: Sounds good to me. I'll rename the PR to WIP, to indicate that it'll change significantly.
Is it OK if we tackle this possible two weeks from now?

patmarion: Oops, didn't realize there is already a GUI class, the `LcmLogPlayerGui` is perfect.  Just pass the `screenGrabberPanel` variable into the ctor of the `LcmLogPlayerGui`.

patmarion: Of course, please feel free to leave open and update me when you want.  Maybe I'll do some work on this, too.  I'm not using LCM in my current project but I am using the screen recorder with other replay mechanisms, and I definitely want to add more features to support programmable video recording.

patmarion:
@EricCousineau-TRI I merged https://github.com/RobotLocomotion/director/pull/522

This PR includes code from that one, and also WIP modifications to screengrabberpanel.py.  Let me know if you plan to update this or whether I should close it?

EricCousineau-TRI:
Howdy Pat, I haven't needed this functionality again yet, so I'll go ahead and close the issue. If I or someone else does need it, I'll consider doing this in a more sound way, and either re-open or reference this commit.
Thanks!

PR #527, opened 2017-06-28
Update director to build with drake/bazel

fbudin69500:


distro/superbuild/cmake/externals.cmake
```diff
@@ -451,7 +460,7 @@ if(USE_PCL AND NOT USE_SYSTEM_PCL)
       -DBUILD_tools:BOOL=OFF
       -DBUILD_apps:BOOL=OFF
       -DBUILD_visualization:BOOL=OFF
-      -DCMAKE_CXX_FLAGS:STRING=-std=c++11
+      "-DCMAKE_CXX_FLAGS:STRING=-std=c++11 ${gxx_extra_flags}"
```
patmarion: what platform was failing here?  Any idea why you encountered a new error here?

src/app/CMakeLists.txt
```diff
@@ -230,7 +230,7 @@ endif()
 set(_install_targets ddApp directorPython drake-visualizer)
 
 add_executable(drake-visualizer drakeVisualizerApp.cpp)
-target_link_libraries(drake-visualizer ddApp)
+target_link_libraries(drake-visualizer ddApp ${PYTHON_LIBRARIES})
```
patmarion: any idea why this was necessary?

cmake/modules/FindLibBot.cmake
```diff
@@ -7,14 +7,21 @@
 # LIBBOT_INCLUDE_DIRS
 # LIBBOT_LIBRARIES
 
+find_package(GTK2 REQUIRED gtk)
+
+find_package(libbot QUIET)
```
patmarion: i think that this new code probably isn't actually being tested on travis-ci?  We do have a libbot build on travis-ci, but it's not compiled by drake.

patmarion:
Thanks! I just posted a few questions, otherwise looks good to me.

src/app/CMakeLists.txt
fbudin69500: It is actually not necessary. Removing it. (I may have had to add it at the beginning because of another issue due to VTK, not 100% sure)

cmake/modules/FindLibBot.cmake
patmarion: Ok.  Yeah I'm not thrilled about some of these recent changes.  I understand why they are necessary changes for you to get things to build with drake bazel.  My concern is that it introduces secondary code paths that are only executed when using a drake bazel build, so this code could break.  I'll get Spartan CI running soon, that will help.

Maybe a good eventual solution would be for the projects like libbot to be updated to produce libbot-config.cmake so then the findpackage can be unified, if that makes sense.

PR #528, opened 2017-06-28
VTK 8.0 and QVTKOpenGLWidget integration

sankhesh:
Supersedes #523.

Implements suggestions from https://github.com/RobotLocomotion/director/pull/523#issuecomment-309072144


patmarion:
this is great, thank you very much!

PR #533, opened 2017-07-06
fix symlink capitalization in mac package

### LINKED ISSUE #532: Capitalization issue in mac packages
This was pointed out in https://github.com/rdeits/DrakeVisualizer.jl/issues/58 but I think it's not Julia-specific. It looks like the packaged libraries are symlinked to a folder named `MacOs/lib`, but that should probably be `MacOS/lib`. The offending line is probably here: https://github.com/RobotLocomotion/director/blob/990c869ae40391135e7dc210c02b151ddf454627/distro/package/make_mac_package.sh#L51

This only shows up on macOS when running with a case-sensitive file system (which is a pretty uncommon configuration).

rdeits:
fixes #532 (I think). I can't actually test this without a case-sensitive file system, but I think this is the right fix.

### DIFF
--- a/distro/package/make_mac_package.sh
+++ b/distro/package/make_mac_package.sh
@@ -48,10 +48,10 @@ cd $scriptDir
 mkdir $packageName
 mv $bundleDir $packageName/
 cd $packageName
-ln -s $appName.app/Contents/MacOs/bin
-ln -s $appName.app/Contents/MacOs/lib
-ln -s $appName.app/Contents/MacOs/include
-ln -s $appName.app/Contents/MacOs/share
+ln -s $appName.app/Contents/MacOS/bin
+ln -s $appName.app/Contents/MacOS/lib
+ln -s $appName.app/Contents/MacOS/include
+ln -s $appName.app/Contents/MacOS/share
 
 # remove headers
 #find $appName.app -name \*.h | xargs rm


patmarion:
BTW---I updated the travis-ci scripts to upload director packages here:

https://bintray.com/patmarion/director/director

I can still manually upload them to http://people.csail.mit.edu/patmarion/software/director/releases/ but maybe you can have a look at the bintry option and provide feedback?

rdeits:
bintray looks fine to me. It looks like I'd just want to update my scripts to grab the binaries from here: https://dl.bintray.com/patmarion/director/ ?

patmarion:
yeah, or here:

https://bintray.com/patmarion/director/director/0.1.0-140-g08e78dc#files

the bintray docs recommend doing this command:

`curl -L "https://dl.bintray.com/patmarion/director/<FILE_PATH>" -o <FILE.EXT>`

patmarion:
here are the latest:

https://bintray.com/patmarion/director/director/0.1.0-144-g7ce9a5a?sort=&order=#files

( i need to sort out the documentation tarballs and automatic deploy of html, that's not currently running automatically)

rdeits:
Cool. Is the upload to bintray automatic on commits to master?

rdeits:
Ah, nvm, just saw the updated files. Thanks again!

patmarion:
the upload is automatic but the files are private until I go to bintray and click the "publish" button.  After I verify it is definitely uploading correctly I plan to set it to publish automatically, too.

PR #537, opened 2017-07-24
VTK8 and MacOS compilation

fbudin69500:


fbudin69500:
@patmarion : This is the follow-up of https://github.com/patmarion/PointCloudLibraryPlugin/pull/2 and https://github.com/RobotLocomotion/director/commit/3b0f33c58753a52549ebb830cf4148d5d0302f3a

patmarion:
Got it, thanks.  Was cmake really doing an infinite recursion on mac case insensitive filesystem (when drake or libbot are enabled)?  I never saw that before.

patmarion:
I will wait for the Travis-CI linux builds to complete and then merge this.  Unfortunately the Travis-CI mac build is failing due to a homebrew change last week.  I need to fix it by pinning a fixed homebrew version until I can update the bottles.

fbudin69500:
@patmarion : Yes, CMake was segfaulting. I understood the problem thanks to this link: https://groups.google.com/forum/#!topic/dealii/AvcXabYxQo0
On Linux it was not a problem as the file systems are (most of the time) case sensitive. I thought it was the same on MacOS, so I did not worry about that with my previous patch.

fbudin69500:
@patmarion : Indeed, the TravisCI MacOS build is failing due to homebrew (numpy install) as you mentioned. The other tests seem to pass.

fbudin69500:
Thanks!

PR #539, opened 2017-07-27
Initialize render window size in ddQVTKWidgetView constructor

fbudin69500:
The render window size is updated when a QResizeEvent happen. At construction
time, the render window size is set to (0,0). This can cause problem if one
requests the size of the render window (e.g. to compute its aspect ratio)
before a QResizeEvent happen.


fbudin69500:
This solves the failing test `test_testDepthScanner`

fbudin69500:
Yes, it was failing with Qt5/VTK8 in the Spartan build

fbudin69500:
@patmarion Ping for my comment ago since you said your notifications were disabled.
What configurations are tested on Travis-CI?

patmarion:
Python 2, Qt 4.8, and vtk 7.1 are tested on Travis-CI.

I don't have notifications disabled.  The only notification I don't get is when you add a commit to a PR.

fbudin69500:
I see. I will keep that in mind.

patmarion:
I'd be happy to have a qt5 and vtk8 configuration to travis-ci, but I haven't had time to set it up yet.

PR #542, opened 2017-08-03
Replace `pydrake.getDrakePath()` with `pydrake.common.FindResourceOrT…

fbudin69500:
…hrow(...)`

`pydrake.getDrakePath()` does not work with the install tree of drake built
with bazel. Instead, we can rely on `pydrake.common.FindResourceOrThrow(...)`
once the environment variable `DRAKE_RESOURCE_ROOT` is defined.

### DIFF
--- a/src/python/director/drcargs.py
+++ b/src/python/director/drcargs.py
@@ -84,18 +84,19 @@ class DRCArgParser(object):
             import pydrake
         except ImportError:
             return False
+        if 'DRAKE_RESOURCE_ROOT' not in os.environ:
+            return False
         return True
 
     def addDrakeConfigShortcuts(self, directorConfig):
 
         import pydrake
-        drakePath = pydrake.getDrakePath()
 
         directorConfig.add_argument(
             '--iiwa-drake',
             dest='directorConfigFile',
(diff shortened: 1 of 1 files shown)


fbudin69500:
This requires https://github.com/RobotLocomotion/drake/pull/6766

fbudin69500:
@patmarion @stonier This is the last requirement to be able to build spartan with drake/bazel

patmarion:
Does this require an environment variable to be defined?  Should we add a check for the environment variable to the function `_isPyDrakeAvailable()` ?

fbudin69500:
The current solution does require an environment variable. I have added a check in the function `isPyDrakeAvailable()` to verify that the environment variable is set.

patmarion:
Cool thanks!

In general I'm not too happy with the current design pattern of the command line args for this.  Basically `--iiwa-drake` is a shortcut for `--director-config=/path/to/drake/examples/kuka_iiwa_arm/director_config.json`.  Which is annoying because it means the application has to resolve that pathname when the command line args are declared, instead of resolving it only when the flag is actually used.

PR #544, opened 2017-08-10
Support building and linking static libraries

jamiesnape:
PTAL @patmarion. FYI @stonier.


patmarion:
Looks fine to me.  Should we add a test configuration on Director's Travis-CI or will it be covered elsewhere?  I'm slightly concerned about maintaining static build support over time but this PR seems pretty minimal so maybe it won't be too bad.

PR #547, opened 2017-08-14
Build position independent code

jamiesnape:
Fixes issue arising from #544. PTAL @patmarion.


patmarion:
im curious, what's the use case for building director with static libs?  is it part of a bazel build system for director?

distro/superbuild/cmake/externals.cmake
```diff
@@ -35,6 +35,7 @@ set(default_cmake_args
   "-DCMAKE_PREFIX_PATH:PATH=${install_prefix};${DRAKE_SUPERBUILD_PREFIX_PATH};${CMAKE_PREFIX_PATH}"
   "-DCMAKE_INSTALL_PREFIX:PATH=${install_prefix}"
   "-DCMAKE_BUILD_TYPE:STRING=${CMAKE_BUILD_TYPE}"
+  "-DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=ON"
```
patmarion: this is a new flag that is always on, so it could change behavior.  i assume it is fine, as most (maybe all?) code compiled during the director superbuild had shared libs enabled.

jamiesnape: Yes, as far as I can tell, everything had shared libs enabled before.

PR #555, opened 2017-09-08
Update cv-utils git tag and pass more cmake flags down to external projects

jamiesnape:
PTAL @patmarion.


patmarion:
Just curious, what's the use case for the new cmake flags?

jamiesnape:
CCache mostly.

patmarion:
Do users set all these flags on the command line?

jamiesnape:
The `CMAKE_<LANG>_COMPILER_LAUNCHER` variables are the most relevant one, and you would set them to the location of CCache or to a launcher script.

PR #557, opened 2017-09-15
Update versions of external projects to match those used by drake

jamiesnape:


patmarion:
Thanks for opening the PR to share your modifications.  Unfortunately I can't accept everything in this PR as it prevents me from maintaining director packages.  Some parts of this PR are fine, like upgrading the versions of eigen and lcm etc, but I can't currently take the parts that modify URLs from which vtk precompiled is fetched or homebrew.

jamiesnape:
>  but I can't currently take the parts that modify URLs from which vtk precompiled is fetched or homebrew.

Why? You can make pull requests like anyone else on the homebrew repo.

jamiesnape:
FYI @stonier. Seems Drake will have to continue to run its own CI on Director after all.

PR #562, opened 2017-09-27
start the mTimer during initialization of ddLCMSubscriber

manuelli:
Previously the `mTimer` wasn't being started. If you then called `setSpeedLimit(60)` then the notification would never get passed up to `python` due to the fact that `mTimer.elapsed()` was always equal to zero. So we would get stuck [here](https://github.com/RobotLocomotion/director/blob/master/src/app/ddLCMSubscriber.h#L162). 

I think this bug has been here all along. Probably what happened is that the default behavior changed when we upgraded the Qt version (or it's simply undefined and behaves strangely on 16.04). Anyways I think this PR fixes the issue.

@patmarion

### DIFF
--- a/src/app/ddLCMSubscriber.h
+++ b/src/app/ddLCMSubscriber.h
@@ -35,6 +35,7 @@ public:
     this->mEmitMessages = true;
     this->mNotifyAllMessages = false;
     this->mRequiredElapsedMilliseconds = 0;
+    this->mTimer.start();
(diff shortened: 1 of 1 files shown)


patmarion:
just tested in qt 4.8.7 and qtime object does start automatically, but the QTime documentation says you should have to call start(), so looks like this is a behavior change that was fixed in Qt5.

PR #563, opened 2017-10-04
Allow toggling of antialiasing of rendering

gizatt:
e.g. [as done here in LabelFusion](https://github.com/RobotLocomotion/LabelFusion/blob/master/scripts/renderTrainingImages.py#L33) for rendering training images

### DIFF
--- a/src/app/ddQVTKWidgetView.cpp
+++ b/src/app/ddQVTKWidgetView.cpp
@@ -110,7 +110,7 @@ ddQVTKWidgetView::ddQVTKWidgetView(QWidget* parent) : ddViewBase(parent)
   this->Internal->VTKWidget->SetRenderWindow(this->Internal->RenderWindow);
 #endif
   this->Internal->VTKWidget->SetRenderWindow(this->Internal->RenderWindow);
-  this->Internal->RenderWindow->SetMultiSamples(8);
+  this->Internal->RenderWindow->SetMultiSamples(0);
   this->Internal->RenderWindow->StereoCapableWindowOn();
   this->Internal->RenderWindow->SetStereoTypeToRedBlue();
   this->Internal->RenderWindow->StereoRenderOff();
--- a/src/app/ddQVTKWidgetView.cpp
+++ b/src/app/ddQVTKWidgetView.cpp
@@ -164,7 +164,7 @@ void ddQVTKWidgetView::init(bool disable_anti_alias)
 
   this->connect(&this->Internal->RenderTimer, SIGNAL(timeout()), SLOT(onRenderTimer()));
   this->Internal->RenderTimer.start();
-  this->setLightKitEnabled(true); 
+  this->setLightKitEnabled(true);
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


peteflorence:
@patmarion hope this one is pretty straightforward, this is just PRing the anti-aliasing feature needed for LabelFusion.

As you may remember, the anti-aliasing has to be set before the GL context is set up, so the only way we could get this to work was inside the cpp constructor.

This PR continues to support the existing Python interface (argument-less) 
```
app = mainwindowapp.construct()
```
and defaults to using anti-aliasing, but also allows 
```
app = mainwindowapp.construct(disable_anti_alias=True)
```

src/app/ddQVTKWidgetView.cpp
```diff
@@ -110,7 +120,11 @@ ddQVTKWidgetView::ddQVTKWidgetView(QWidget* parent) : ddViewBase(parent)
   this->Internal->VTKWidget->SetRenderWindow(this->Internal->RenderWindow);
 #endif
   this->Internal->VTKWidget->SetRenderWindow(this->Internal->RenderWindow);
```
peteflorence: @sankhesh do you agree we can use only the second call the SetRenderWindow?  this comes from this commit: https://github.com/RobotLocomotion/director/commit/dd4ee5d12ebbf505279d4248f2530ce5fc87c3b4#diff-ab3b6e434c7408392ed6672f3ef053d8

sankhesh: @peteflorence Thanks for spotting this. I think that is just oversight. It would be fine to remove the `SetRenderWindow` call within the `#else` block.

peteflorence: :+1: thanks

patmarion:
Looks fine, but, what about using a static cpp variable to globally set rendering options instead of passing the options all the way through the mainwindowapp python code.  It seems reasonable to me to apply these settings application wide, rather than per-view as constructor arguments to the ddQVTKWidgetView.


So user code could look like this:

```
# before constructing the first view, set some rendering options:
import director.visualization as vis
vis.disableAntiAliasing()

# now create the app and view
app = mainwindowapp.construct()
```


And the implementation of `vis.disableAntiAliasing()` would look like:

```
def disableAntiAliasing:
  PythonQt.dd.ddQVTKWidgetView.setAntiAliasing(False)
```

And then create a static cpp function:

```
class ddQVTKWidget {
...

  static void setAntiAliasing(bool enabled);
};
```

And in the cpp:

```
namespace {
  antiAliasingEnabled = True
};

void ddQVTKWidgetView::setAntiAliasing(bool enabled)
{
  antiAliasingEnabeld = enabled
}
```

patmarion:
Also, style fix:

instead of `disable_anti_aliasing` please us `disableAntiAliasing` to be consistent with the current style

peteflorence:
Pat good idea, that is cleaner.   Not sure when can get to it but I can try out implementing that way

manuelli:
Superseded by #611.

PR #565, opened 2017-10-04
Update bot image queue for pointcloud visualization

gizatt:
- Enables build of bot_image_queue
- Disables kinectLCM driver
- Updates cameraview.py to be compatible with VTK \geq 6

Submitting so testing can spin up, but I'm happy to add a build option to allow this bit of code to only build when requested, since it has a more involved dependency list.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,7 @@ option(USE_DRC_MAPS "Build director with drc maps dependency." OFF)
 option(USE_DRC_PLANE_SEG "Build director with drc plane segmentation." OFF)
 option(USE_OCTOMAP "Build director with octomap dependency." OFF)
 option(USE_COLLECTIONS "Build director with collections dependency." OFF)
+option(USE_PERCEPTION "Build director features that require OpenCV, PCL, cv-utils, and libbot as dependencies." OFF)
 
 use_cpp11()
 setup_qt()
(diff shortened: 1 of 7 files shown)


gizatt:
Tweaked to build the new stuff only when the dependencies are available

peteflorence:
We should test this on VTK 6/7 before merging (so far only tested on VTK 8)

patmarion:
I can merge all the changes except the changes to the CMakeLists.txt.  For the CMakeLists.txt, I'd like to introduce the option "USE_PERCEPTION" which enables the search for opencv, pcl, and cv-utils and adds ddBotImageQueue to the build.  I would also keep the USE_DRC option in case there are still some openhumanoids specific requirements.

In general I like to be explicit, by having options like USE_PERCEPTION, rather than searching for PCL and if it is found then automatically enabling files.  Because then you can't turn things off except by uninstalling PCL.

patmarion:
> We should test this on VTK 6/7 before merging (so far only tested on VTK 8)

Thanks for the suggestions.  I think the vtk api changes look reasonable to me and are likely to work on vtk7.

gizatt:
Thanks for the suggestion, switched to a USE_PERCEPTION flag. Did I put that option everywhere I ought to?

patmarion:
LGTM!  Thanks.  I will wait for Travis-CI to post the results and then merge.

PR #571, opened 2017-11-06
Adds support for mesh geometries to SDF scene loader

ferrolho:
This pull request adds support for loading mesh geometries from SDF files and also adds support for scaling of meshes.

Maybe this commit can also be cherry-picked onto master after merge?

cc: @wxmerkt

### DIFF
--- a/src/python/director/affordanceitems.py
+++ b/src/python/director/affordanceitems.py
@@ -241,15 +241,16 @@ class MeshAffordanceItem(AffordanceItem):
             if os.path.isfile(filename):
                 polyData = ioUtils.readPolyData(filename)
 
-                transform = vtk.vtkTransform()
-                transform.Scale(scale)
+                if not scale == [1, 1, 1]:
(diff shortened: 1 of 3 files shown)


src/python/director/affordanceitems.py
```diff
@@ -223,19 +227,29 @@ def __init__(self, name, view):
...
             if not os.path.isabs(filename):
                 filename = os.path.join(director.getDRCBaseDir(), filename)
 
             if os.path.isfile(filename):
                 polyData = ioUtils.readPolyData(filename)
+
+                transform = vtk.vtkTransform()
+                transform.Scale(scale)
+
+                transformFilter = vtk.vtkTransformPolyDataFilter()
```
patmarion: im not that worried about it, but possible speed up you can skip the filter if scale==[1.0, 1.0, 1.0]

patmarion:
do you want to open a PR against robotlocomotion/master too?

PR #575, opened 2017-11-09
In drake, convexHull.h was renamed to convex_hull.h and moved

fbudin69500:
This change happened in drake commit c40f4a04d211934d553e99f0e5a164f4b0f806c5.

The commit message states:
    Move convex_hull from /util to /systems/robotInterfaces

    It's only used from within controllers, and shouldn't be used elsewhere.
    We mark it private, since it's only used from .cc files, also.


fbudin69500:
@patmarion

patmarion:
sorry i completely missed this!  i thought i had already clicked merge.

PR #580, opened 2018-01-16
update sha1 of QtPropertyBrowser used for qt5 build

patmarion:
This pulls a version of QtPropertyBrowser where extra
compiler flags like -march-native have been removed.

See:

https://github.com/RobotLocomotion/drake/issues/7740


patmarion:
unfortunately none of the travis-ci builds actually enable qt5, so there's nothing to see here...

also travis-ci has been failing because the cdash server it submits to if offline :(  need to setup a new one.

PR #583, opened 2018-01-26
Document `triad` in remote tree viewer

rdeits:


patmarion:
btw, i noticed that the Python client for remote tree viewer doesn't have pointcloud support.  I added it in a file I was writing, but I'd like to add it to director, too.

PR #584, opened 2018-01-26
support PointCloud from the python tree viewer client

rdeits:
@patmarion as you wish :wink:

### DIFF
--- a/src/python/director/viewerclient.py
+++ b/src/python/director/viewerclient.py
@@ -125,6 +125,20 @@ class Triad(BaseGeometry):
             "tube": self.tube
         }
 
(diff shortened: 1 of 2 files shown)


patmarion:
hooray!  i thought there was a chance my comment could generate a PR ;)


sorry for the noisy travis-ci results.  travis-ci is set to upload test results to one of the MITDRC laptops in RLG, but that server is offline now.  This is where the IP address is defined:

https://github.com/RobotLocomotion/director/blob/master/CTestConfig.cmake#L5

I think it was drc-001 laptop.  It has a script that starts the cdash server inside a docker instance.

PR #587, opened 2018-01-31
add taskrunner.py and test

patmarion:


patmarion:
This adds the TaskRunner class to director.  This is the same class that was added to Spartan last year, but with one bugfix: 

https://github.com/RobotLocomotion/spartan/blob/master/modules/spartan/utils/taskrunner.py

PR #594, opened 2018-02-16
Make pydrake optional in drakevisualizer.py

rdeits:
This makes it possible for future versions of drake to completely remove `drake-visualizer`'s dependency on pydrake. It should be fully backwards compatible (since it just prevents package lookup from crashing if pydrake is not found).


rdeits:
cc @jwnimmer-tri

patmarion:
Sorry I missed this while I was out of town.

PR #595, opened 2018-02-20
Fix deprecated usages of RigidBodyTree::bodies

EricCousineau-TRI:
See https://github.com/robotlocomotion/drake/pull/8085 for more info.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/director/595)


src/app/ddDrakeModelOH.cpp
```diff
@@ -445,12 +445,12 @@ class URDFRigidBodyTreeVTK : public RigidBodyTree
 
     RigidBodyTree* model = this;
 
-    const std::shared_ptr<RigidBody> worldBody = model->bodies[0];
```
wxmerkt: I don't think the OH model should be touched at all - it uses an older, fixed version of Drake and thus should not be updated if the upstream API changes. Only the ddDrakeModel should be updated.

EricCousineau-TRI: Done. Sorry about that!

PR #596, opened 2018-03-20
Update to pcl 1.8.1 and newer PointCloudLibraryPlugin

sammy-tri:
This allows compilation to succeed with newer versions of VTK which
remove some deprecated functions.


patmarion:
Thanks @sammy-tri!

PR #597, opened 2018-03-21
Various compatibility updates

jamiesnape:
Relates https://github.com/RobotLocomotion/spartan/issues/266. PTAL @patmarion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,11 @@ cmake_minimum_required(VERSION 2.8.7)
 # Drake Designer
 project(DD)
 
+if(POLICY CMP0028)
+  # Double colon in target name means ALIAS or IMPORTED target.
+  cmake_policy(SET CMP0028 NEW)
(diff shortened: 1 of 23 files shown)


patmarion:
Thanks @jamiesnape.  Looks like this PR failed the ubuntu14 tests.  While I would be happy to drop Ubuntu 14 at this point, I think it is easy to fix this failure by modifying `externals.cmake`.  See this line:

https://github.com/RobotLocomotion/director/blob/master/distro/superbuild/cmake/externals.cmake#L154

We need a similar line here:

https://github.com/RobotLocomotion/director/blob/master/distro/superbuild/cmake/externals.cmake#L583

Basically what is happening, is for systems without cmake < v3.1, we download a new version of cmake and use it when configuring projects that use lcm.

patmarion:
i made the required fix in `master` so this PR would pass on travis-ci ubuntu14.  thanks!

PR #606, opened 2018-06-08
Searches LCM first using CMake and then using pkgconfig

fbudin69500:
LCM can generate a lcm-config.cmake file which can be used to find
LCM on the system. Instead of requiring LCM to be found with
pkgconfig, this patch modifies director to first search for LCM
using CMake `find_package()` and if this does not work, fall back
on the original implementation using pkgconfig to find the necessary
files.


patmarion:
thanks, will wait for Travis-ci then merge

fbudin69500:
Nevermind, I just realized I was working on an old branch. This problem has already been solved by @jamiesnape in #597 .

PR #608, opened 2018-06-20
remove unnecessary line causing bug

### LINKED ISSUE #607: error using `start ik` button in kuka director app
When you hit `start ik` you get the following error

```
>>> Traceback (most recent call last):
  File "/home/manuelli/spartan/build/install/lib/python2.7/site-packages/director/teleoppanel.py", line 1004, in startIk
    om.removeFromObjectModel('reach goal')
  File "/home/manuelli/spartan/build/install/lib/python2.7/site-packages/director/objectmodel.py", line 530, in removeFromObjectModel
    _t.removeFromObjectModel(obj)
  File "/home/manuelli/spartan/build/install/lib/python2.7/site-packages/director/objectmodel.py", line 337, in removeFromObjectModel
    self._removeItemFromObjectModel(self._getItemForObject(obj))
  File "/home/manuelli/spartan/build/install/lib/python2.7/site-packages/director/objectmodel.py", line 241, in _getItemForObject
    return self._objectToItem[obj]
KeyError: 'reach goal'
```

it is due to trying to remove `reach goal` when it does not exist. I think an easy solution is to check if `reach goal` exists, and remove it if it does. Alternatively we can just use `vis.updateFrame` instead of `vis.showFrame` [here](https://github.com/RobotLocomotion/director/blob/master/src/python/director/teleoppanel.py#L1005) and drop line [1004](https://github.com/RobotLocomotion/director/blob/master/src/python/director/teleoppanel.py#L1004)

manuelli:
fixes #607

### DIFF
--- a/src/python/director/teleoppanel.py
+++ b/src/python/director/teleoppanel.py
@@ -1001,7 +1001,7 @@ class GeneralEndEffectorTeleopPanel(object):
             graspToWorld.Concatenate(endEffectorLinkFrame)
             reachGoal = graspToWorld
 
-        om.removeFromObjectModel('reach goal')
+        
         goalFrame = vis.showFrame(reachGoal, 'reach goal', scale=0.1, parent=EndEffectorTeleopPanel.getConstraintFrameFolder())
         goalFrame.setProperty('Edit', True)
 


patmarion:
thanks!

PR #614, opened 2019-01-16
Fix camera zoom and centering with VTK 8

manuelli:
Fixes the issue described in RobotLocomotion/drake#8940. Thanks to @EricCousineau-TRI for the fix.


manuelli:
Closing due to this being a workaround rather than a fix.

src/python/director/affordanceitems.py
```diff
@@ -246,7 +246,7 @@ def updateGeometryFromProperties(self):
                     transform.Scale(scale)
 
                     transformFilter = vtk.vtkTransformPolyDataFilter()
-                    transformFilter.SetInput(polyData)
+                    transformFilter.SetInputData(polyData)
```
patmarion: this one can definitely land

PR #617, opened 2019-03-04
fix bug in crop to bounds

manuelli:
Previously `zip` was only traversing X and Y axes, not Z.


patmarion:
maybe just the docstribg is wrong?  It works if bounds is a 3x2 instead of a 2x3?

manuelli:
Yeah that's probably right. The only other usage is in [drilldemo](https://github.com/RobotLocomotion/director/blob/master/src/python/director/drilldemo.py#L2396) and they pass in a list of lists. The original list is length 3. So yeah seems like just an update on the documentation is needed.

